PR #12, opened 2022-10-22
[prismlauncher] improve spec readability and compatibility

getchoo:
missed an underscore in my previous macro for qt5 dependencies on opensuse, so now it works :)

also added support for rhel and centos being the build platform, as well as added some spacing so the endless if statements don't look too messy

this now mirrors what is on the repo i use for copr (or will when i push it there anyways), as well as the OBS, so it'll make my life a little easier. leaves the door open for building copr packages directly from this repo too


lleyton:
building the copr package from this repo isn't a bad idea ;)

besides that, LGTM

PR #15, opened 2022-10-27
add prismlauncher qt5

getchoo:
[build succeeded on copr](https://copr.fedorainfracloud.org/coprs/g3tchoo/prismlauncher/build/4993851/)

i also have nightly packages up, but im going to talk to cappy about how chkupdate.py could be used for commits instead of releases, after that i'll merge them as well - and probably start building the copr packages from here because dear god i don't want to run my own script manually every night and stuff

### DIFF
--- a/anda/games/prismlauncher-qt5/anda.hcl
+++ b/anda/games/prismlauncher-qt5/anda.hcl
@@ -1,0 +1,6 @@
+project "pkg" {
+    rpm {
+        spec = "prismlauncher-qt5.spec"
+        sources = "."
+    }
+}
--- a/anda/games/prismlauncher-qt5/change-jars-path.patch
+++ b/anda/games/prismlauncher-qt5/change-jars-path.patch
@@ -1,0 +1,26 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 48411175..5bd7e0dd 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -250,7 +250,7 @@ if(UNIX AND APPLE)
+ elseif(UNIX)
+     set(BINARY_DEST_DIR "bin")
+     set(LIBRARY_DEST_DIR "lib${LIB_SUFFIX}")
+-    set(JARS_DEST_DIR "share/jars")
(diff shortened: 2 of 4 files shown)


windowsboy111:
chkupdate once looked up the commits instead of the release though cappy told me having it looking for releases is better off 
see commit history

windowsboy111:
https://github.com/terrapkg/packages/blob/3619f9d4a6d371f90b478eb0c902ca989c54ebc0/anda/prismlauncher/chkupdate.py

Checkout this version of chkupdate

korewaChino:
> chkupdate once looked up the commits instead of the release though cappy told me having it looking for releases is better off see commit history

this was for nightly builds. we only changed it because prism went stable

lleyton:
Maybe we could have a separate -git package which updates on commit?

getchoo:
> Maybe we could have a separate -git package which updates on commit?

i was thinking of this, but that would put a lot of stress on copr/github actions with builds most likely triggering after every single run - so i think nightly is a better solution.

i also have a script working now (thanks for the tip btw @windowsboy111, i took some of the stuff from the older chkupdate.py) but i'm not really sure how updates could be limited to once per day. i was thinking of using [sched](https://docs.python.org/3/library/sched.html) with the delayfunc set to 24h, but that would just make the action last longer than needed and probably time out at some point. i also considered just grabbing the system time and comparing it to a set time, but then it might not fall within the time the workflow runs - unless one of you know a certain hour the workflow does run every day, then it's probably possible. i think the best solution to this would probably be a separate workflow for nightly packages. it'd just be a lot more streamlined and not require any weird time magic in the scripts themselves that could cause issues in some random edge case down the line.

this is also ready for merge now, i can just put the regular nightly and qt5-nightly packages into a different PR once we get this sorted out

windowsboy111:
The workflow is basically a [cron job](https://github.com/terrapkg/packages/blob/main/.github/workflows/update.yml#L7) that runs every two hours.
So maybe you can check if it's at hour 0
```py
from datetime import datetime
if datetime.utcnow().hour != 0: exit(f'{NAME}: skipping update check')
```

windowsboy111:
I'll change the schedule later, not really that urgent for now

PR #17, opened 2022-10-28
[primslauncher-nightly] epic comment fail

getchoo:
welp...forgot to uncomment the actually important line after testing the script :/


windowsboy111:
ahhhhhh lol, it's ok
we make mistakes all the time and yeah we just make a 100 commits until it works

windowsboy111:
ahhhhhh lol, it's ok
we make mistakes all the time and yeah we just make a 100 commits until it works

PR #18, opened 2022-11-01
remove unneeded patch for prismlauncher

getchoo:
the patch we were using has been [upstreamed](https://github.com/PrismLauncher/PrismLauncher/pull/218) and included in the 5.1 release


lleyton:
yummy

PR #22, opened 2022-11-07
[prismlauncher] update conflicts/make specs more consistent

getchoo:
the only thing affecting packaging here is adding proper conflicts so that none of the packages can be installed together. this might change later with a patch since it was suggested in the prism discord, but i'm a little lazy

the rest is just making the specs as consistent as possible between each other, just so they all look nice :p


getchoo:
also write access would be cool if you guys are comfortable with it. it would make quick fixes like this a lot easier and i wouldn't have to wait to rebuild on copr (no pressure /gen)

lleyton:
I'd love to give you access, we just gotta discuss internally first :)

besides that, LGTM!

PR #23, opened 2022-11-08
fix conflicts (again)

getchoo:
missed an underscore


getchoo:
might have found a better way to end this (surprised i missed it if it works), so making this a draft for now

getchoo:
never mind, the issue was from older builds before my previous commit. this just fixes the macro used for conflicts (and alignment between tags that i missed and annoyed me)

PR #34, opened 2022-11-24
[prismlauncher-nightly] bump commit

getchoo:
not really sure what's going on, but it looks like auto updates are no longer working for the nightly packages. since the script still works for me locally and each check after discord-ptb fails, i'm guessing maybe there's some rate limiting going on here. is there any way to solve this?


lleyton:
The rate limiting issue should be fixed now in https://github.com/terrapkg/packages/commit/3a81bc0cc02c7e521c599566ba4dea2a7d3f209b

Issue #26, opened 2022-11-11
Add chkupdate scripts for all packages


lleyton:
Currently, we have quite a few packages without `chkupdate.py` scripts. This has led some of the older packages, such as those ported from tauOS, to fall behind. (See https://repology.org/projects/?inrepo=terra_37&outdated=1)

windowsboy111:
I modified `update.sh` to print out dirs without chkupdate.py:
```sh
search() {
    for folder in $1/*; do
        if [ -f "$folder/chkupdate.py" ]; then
            continue
        else
            echo $folder
        fi
        x=0
        for thing in $folder/*; do
            [[ -f $thing ]] && x=1 && break
        done
        [[ $x -eq 0 ]] && search $folder
    done
}

search anda
```

Currently by 1fff42c, the following dirs do not have chkupdate.py:
```
anda/adwaita++-icons
anda/apps
anda/apps/openasar-canary
anda/audio
anda/audio/distrho
anda/dart
anda/desktops
anda/desktops/budgie
anda/desktops/budgie/budgie-extras
anda/desktops/elementary
anda/desktops/elementary/evolution-data-server
anda/desktops/elementary/pantheon-session-settings
anda/desktops/elementary/plank
anda/dnf-plugins-core
anda/fonts
anda/fonts/arphic-ukai
anda/fonts/arphic-uming
anda/fonts/hannom
anda/fonts/kanjistrokeorders
anda/fonts/mini-wakuwaku
anda/fonts/seto
anda/games
anda/games/minecraft-java
anda/go
anda/go/curlie
anda/lib
anda/lib/libappimage
anda/lib/marble-gtk
anda/lib/poppler
anda/moby-extras
anda/moby-extras/moby-buildx
anda/moby-extras/moby-compose
anda/mock-configs
anda/rust
anda/rust/bat-extras
anda/rust/pop-launcher
anda/rust/sccache
anda/rust/youki
anda/rust/zellij
anda/srpm-macros
anda/tools
anda/tools/appimagelauncher
anda/tools/electron
anda/tools/gcm-core
anda/tools/lorax
anda/tools/melody
anda/tools/muon
```

madonuko:
# Packages without autoupdate
Update: from #145, we have (update.rhai)
```
anda/audio/distrho
anda/desktops/elementary/pantheon-session-settings
anda/fonts/arphic-ukai
anda/fonts/arphic-uming
anda/fonts/hannom
anda/fonts/kanjistrokeorders
[...]

PR #55, opened 2022-12-05
add chkupdate.py to more packages

getchoo:
adds a chkupdate.py script to 12 out of the 33 (applicable by my count :p) packages listed in #26

maybe i'll work on some others later, these were just the easiest to add tonight

### DIFF
--- a/anda/apps/openasar-canary/chkupdate.py
+++ b/anda/apps/openasar-canary/chkupdate.py
@@ -21,3 +21,10 @@ cur = matches[0]
 if sha == cur:
     exit(f"{NAME}: Up to date!")
 print(f"{NAME}: {cur} -> {sha}")
+
+newspec = re.sub(REGEX_SHA, f"%define commit {sha}", txt)
+
+f.close()
+f = open(SPEC, "w")
+f.write(newspec)
+f.close()
--- a/anda/apps/openasar-canary/chkupdate.py
+++ b/anda/apps/openasar-canary/chkupdate.py
@@ -1,0 +1,23 @@
+import requests
+import re
+import os
+
+
+NAME = "openasar-canary"
+REPO = "GooseMod/OpenAsar"
+REGEX_SHA = r"%define commit (.+)"
+SPEC = f"{NAME}.spec"
+LINK = f"https://api.github.com/repos/{REPO}/commits/HEAD"
+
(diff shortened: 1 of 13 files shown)


lleyton (dismissed):

anda/apps/openasar-canary/chkupdate.py
```diff
@@ -0,0 +1,23 @@
...
+f = open(SPEC, "r")
+
+matches = re.findall(REGEX_SHA, txt := f.read())
+if not len(matches):
+    exit(f"{NAME}: Failed to match regex!")
+cur = matches[0]
+
+if sha == cur:
+    exit(f"{NAME}: Up to date!")
+print(f"{NAME}: {cur} -> {sha}")
```
windowsboy111, line 23: so this file writes nothing?

windowsboy111 (changes requested):
overall the requests.get() thingy looks ugly but meh

anda/apps/openasar-canary/chkupdate.py
getchoo: yep, the version is already determined with [%{autogitdate}](https://github.com/terrapkg/packages/blob/44cf5b5a477d3100daee78cca28c344a685fe750/anda/apps/openasar-canary/openasar-canary.spec#L4) and the [source](https://github.com/terrapkg/packages/blob/44cf5b5a477d3100daee78cca28c344a685fe750/anda/apps/openasar-canary/openasar-canary.spec#L10) just points to the continuously updated nightly release upstream. the goal of this script was to just trigger a rebuild on each new commit (which should signify a new nightly release), which is why i started defining the commit in the spec file

getchoo:
> overall the requests.get() thingy looks ugly but meh

that's just how black formats it. i could move the string slices to a different line but it would be kinda pointless

anda/apps/openasar-canary/chkupdate.py
windowsboy111: I don't think builds will trigger with no files changed in the dir…

getchoo: 🤦🏻 i see what you mean now. i'll push another commit now

...maybe i shouldn't be doing this at 4am

windowsboy111 (approved):

lleyton (approved):

PR #129, opened 2023-01-15
[prismlauncher(-qt5)-nightly] add patch to find cmark correctly

getchoo:
took me a while to figure this out, but it looks like fedora's `cmark-devel` package is at version that doesn't support cmake properly. this adds a patch that will remedy this issue with a new cmake config file sourced from [neochat](https://invent.kde.org/network/neochat/-/blob/3f8d2a11d0e00e35a743996b7f3e71c9a401de81/cmake/Findcmark.cmake)


lleyton (dismissed):

lleyton:
Please amend your prior commits so that they are signed

getchoo:
looks like when i rebased in github it removed the signature lol

lleyton (approved):

PR #135, opened 2023-01-19
Fix nerd-fonts

windowsboy111:


windowsboy111:
https://github.com/terrapkg/packages/actions/runs/3957844576/jobs/6778705388#step:4:501

Fails due to subatomic limitations (?)

lleyton (approved):

PR #145, opened 2023-01-24
Migrate autoupdate to meet anda 0.1.9

### LINKED ISSUE #59: dedicated workflow for nightly packages
with the website and wiki have been updated to advertise the nightly packages in this repo, i think it might be a good time to make a dedicated workflow for nightly packages. 

currently, nightly packages (namely `prismlauncher-nightly`, `prismlauncher-qt5-nightly`, and `openasar-canary`) check for a commit every 30 minutes like the regular packages. imo, this is a bad approach as it misses the point of the packages being considered "nightly," puts an unneeded load on gh actions, increases the likelihood of an unstable commit being pushed to users, and pushes more updates to them than they would probably ever use. a better solution here could be to use the current update.yml workflow but schedule it for every 12-24h for the nightly packages, as that would solve the aforementioned issues, as well as help users viewing the actions tab track nightly updates easier. it would also help mitigate a potential issue with `openasar-canary`'s chkupdate script (as the source we use isn't instantly updated when a new commit is pushed) and not bring up the previous issues found in the nightly prismlauncher packages' approach.

windowsboy111:
close #59


korewaChino (dismissed):

korewaChino (approved):

lleyton (approved):

lainsce (approved):
On the pkgs I am responsible for, lgtm.

PR #152, opened 2023-01-27
Fix switchboard-plug-about

lleyton:


lainsce (approved):
simple and effective :)

PR #154, opened 2023-01-27
Remove checks for elementary-code

windowsboy111:


lainsce (approved):
Yup it already does this while building, so LGTM.

PR #161, opened 2023-02-03
[prismlauncher-nightly] disable tests and explicitly require cmark

getchoo:
tests currently fail on x64 for unknown reasons. this isn't reproducible on upstream's ci as well as building locally. not sure what's going on so it's better to just disable them for now since it seems to be a false positive.

also made cmark an explicit `requires` due to [this](https://bugzilla.redhat.com/show_bug.cgi?id=2166815) bug. i'll revert it once the change makes it way to f38


windowsboy111 (approved):

windowsboy111:
Just realised that doesn't patch the non-nightly version. Maybe I should've been patient since there was literally a new release

PR #258, opened 2023-03-01
fix(files): granite-7

windowsboy111:


rmnscnce (approved):
Did the previous commit even build? Otherwise this patch itself LGTM. Will approve.

PR #282, opened 2023-03-17
feat: add apparmor

cat-master21:


windowsboy111 (approved):
Maybe update.rhai if you can but I can add it later sooo

PR #284, opened 2023-03-17
fix(qt5-pim): fix installing headers

cat-master21:


windowsboy111:
Can you use `%{version}` in `Source0`?

cat-master21:
> Can you use `%{version}` in `Source0`?

No, as it relies on %commit which is part of the forgemeta and then part of the release number (I think) also.

lleyton (approved):

windowsboy111 (approved):

PR #289, opened 2023-03-19
fix: iosevka-fusion-fonts blackbox-terminal

windowsboy111:


windowsboy111:
for iosevka it still doesn't build but at least the deps are fixed

windowsboy111:
nvm fixed

lleyton (approved):

PR #296, opened 2023-03-19
add: python3-rich

madonuko:


windowsboy111:
added because of `melody`

lleyton:
Isn't this being shipped upstream already: https://packages.fedoraproject.org/pkgs/python-rich/python3-rich/

windowsboy111:
upstream is solving outdated deps

PR #314, opened 2023-03-24
feat: add gsettings-qt

cat-master21:


anda/lib/gsettings-qt/gsettings-qt.spec
```diff
@@ -0,0 +1,63 @@
+%global forgeurl https://gitlab.com/ubports/development/core/gsettings-qt
+%global commit d5e002d7e0bce46c315bcc99a44a8bd51f49f488
+%forgemeta
+
+Name:       gsettings-qt
+Version:    0.3
```
windowsboy111, line 6: Forgot to add comment, same issue: there's no 0.3 in gitlab but only 0.2, and I've searched repology and none is on 0.3;

However, I do have to say it's quite outdated so I think maybe we can just straight up add the commit to the version (have seen other repos done that)

windowsboy111:
[v0.2 is not the same as v0.3](https://gitlab.com/ubports/development/core/gsettings-qt/-/compare/v0.2...main?from_project_id=10070038&straight=false), so change commit?
or you can just use `0.3^%commit` as the version

windowsboy111 (approved):

PR #315, opened 2023-03-24
feat: add process-cpp

cat-master21:


anda/lib/process-cpp/process-cpp.spec
```diff
@@ -0,0 +1,84 @@
+%global forgeurl https://gitlab.com/ubports/development/core/lib-cpp/process-cpp
+%global commit ee6d99a3278343f5fdcec7ed3dad38763e257310
+%forgemeta
+
+Name:          process-cpp
+Version:       3.0.2
```
windowsboy111, line 6: `3.0.2` is unreleased. I'm not sure if we should do it like this… Is is a must to use the latest commit?

windowsboy111 (commented):
Auto update script:
```ts
let o = get("https://gitlab.com/api/v4/projects/16508826/repository/tags").json_arr()[0];
rpm.global("commit", o.commit.id);
rpm.version(o.name);
```

anda/lib/process-cpp/process-cpp.spec
cat-master21: My bad it looked like `3.0.2` was released https://gitlab.com/ubports/development/core/lib-cpp/process-cpp/-/blob/main/debian/changelog

windowsboy111: I mean it does say
```
process-cpp (3.0.2) UNRELEASED; urgency=medium
```
unreleased in caps right there… so…

anda/lib/process-cpp/process-cpp.spec
```diff
@@ -0,0 +1,84 @@
+%global forgeurl https://gitlab.com/ubports/development/core/lib-cpp/process-cpp
+%global commit ee6d99a3278343f5fdcec7ed3dad38763e257310
```
windowsboy111, line 2: I think same issue here unfortunately, wrong commit

cat-master21: Meaning what? There are some changes/commits not in `3.0.1`.

windowsboy111: Hmm, maybe just name the commit `3.0.2^%commit`

windowsboy111 (approved):

PR #318, opened 2023-03-25
upd(qt5-qtdbustest): `0.2` -> `0.3`

cat-master21:

PR #319, opened 2023-03-25
Feat: add qtdbusmock

cat-master21:
Depends on #318 first.


windowsboy111 (approved):

cat-master21:
@windowsboy111 #318 wasn't merged first, so when the version is updated `qtdbusmock` will find different symbols so qtdbusmock will need to be rebuilt after.

PR #320, opened 2023-03-25
fix(ci): sync

windowsboy111:


rmnscnce (approved):
You can use the `-s` flag to replace `--signoff` on the `git` command but otherwise it's alright.

lleyton (approved):

PR #345, opened 2023-04-01
fix(lomiri-schemas): revert remove feature

cat-master21:
For some reason, in my personal computer (Fedora 37), the schemas compile fine without the patch and with it fail but in my VM (also Fedora 37) it fails without the patch (opposite of my computer). So **Please** test this before merge.


windowsboy111:
test build runs fine

windowsboy111 (approved):

cat-master21:
> test build runs fine

It's not if it builds but can `/usr/share/glib-2.0/schemas/com.lomiri.Shell.gschema.xml` compile. Please run `glib-compile-schemas /usr/share/glib-2.0/schemas` (or add sudo to save the compiled schemas) and see if `com.lomiri.Shell.gschema.xml` has an error @windowsboy111.

PR #362, opened 2023-04-07
fix(ci): mg

windowsboy111:


windowsboy111:
Found a bug. file name doesn't always contain the actual rpm package name.

lleyton (approved):

PR #368, opened 2023-04-09
chore(nux): change binary diff to source

cat-master21:
Should remove the complaint from linter that the patch isn't applied even though it is.


windowsboy111 (approved):

windowsboy111:
that's pretty smart

PR #369, opened 2023-04-09
chore(libindicator): remove linting warning

cat-master21:
Should hopefully work.


cat-master21:
Nope that didn't work I'll fix it at some point.

windowsboy111 (approved):

PR #378, opened 2023-04-15
feat: add ayatana-indicator-sound

cat-master21:
Note I have purpose-fully added a seperate `indicators/` directory to hold all of the Ayatana Indicators and libindicator. I will move them in a seperate PR.


windowsboy111:
Do you think we can autoupdate this?

cat-master21:
I think I can but would rather not do it automaticlly (but definitely can).

windowsboy111 (approved):

PR #420, opened 2023-05-06
chore: lint

madonuko:


madonuko:
vala-language-server-nightly wontfix

PR #441, opened 2023-05-13
fix: update vala-nightly to depend on libvala-nightly instead of stable

lleyton:


lleyton:
force merging, since this is a critical issue

PR #444, opened 2023-05-14
fix(unitycore): fix linkage

cat-master21:
Not sure why, but it appears that libdir was broken.


lleyton (dismissed):

lleyton:
Hm, I'm going to build a new Terra builder image, hopefully it fixes that weird rpmfusion gpg error.

lleyton:
Weird... so it's only happening on this package (even on the version on main). It's probably pulling something from rpmfusion.. is that intensional?

cat-master21:
No it should not be doing so, I built this package without RPMFusion and without others (like electron).

lleyton:
Alrightie, we decided to get rid of rpmfusion from the Mock build environment, it was causing issues with some other packages. We'll try a run again once the new image is done building.

lleyton:
Alrightie, could you check the latest build?

madonuko (approved):

PR #446, opened 2023-05-18
Rename anda-mock-configs to terra-mock-configs

lleyton:
This PR renames anda-mock-configs to terra-mock-configs, to get rid of the legacy andaman-common naming. The prior package is obsoleted and provided, and symlinks are made for the old mock config names, so packagers shouldn't notice this. We also get rid of rpmfusion, which has been causing build issues.


lleyton:
Once this is merged, I'll update our docs.

madonuko (approved):

PR #447, opened 2023-05-18
Fix wingpanel

lleyton:
We're just pulling in all of the commits from the current release version to the currently latest commit. None of them should be breaking, the only changes are translation and the mutter fixes.


lleyton:
Depends on #446

madonuko (approved):

PR #469, opened 2023-05-25
Feat: add golang-github-fyshos-fynedesk

cat-master21:
I've been expirementing with desktops lately,and there are many more that I plan on doing/trying. Though this desktop doesn't use KDE in any  way.


cat-master21:
Why is git not installing, ugh.

madonuko (approved):

PR #474, opened 2023-05-27
fix(provides): marble-gtk

madonuko:


madonuko:
Closing in favour of #481

Seems like rpm hardcoded the pkg-config which makes the workaround impossible

PR #481, opened 2023-05-29
fix: blackbox-terminal

madonuko:


korewaChino:
now that #650 is done, we can finish this PR

korewaChino (approved):

PR #489, opened 2023-06-01
feat: rm vl-gothic hyperfine

madonuko:
vl-gothic-fonts, vl-gothic-fonts-all, vl-pgothic-fonts, hyperfine are redundant as they are already provided by the Fedora repos. They are consequently removed.


nothingneko (approved):
LGTM

PR #491, opened 2023-06-01
fix: system76-scheduler

madonuko:


lleyton:
Macros are still being expanded, which is probably causing a build failure.

lleyton (changes requested):
Shouldn't we be using Fedora's systemd macros?
https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd

lleyton (approved):

lleyton (approved):

PR #506, opened 2023-06-05
remove libunity

cat-master21:
So after looking at it, Fedora packages it the same as I do so no need for it here.


cat-master21:
This should be the last PR to rename/remove duplicates for me.

madonuko (approved):

PR #510, opened 2023-06-08
prismlauncher: specify jdk 17 + cleanup outdated patches/scriptlets

getchoo:
with prismlauncher 7.0 releasing soon™️ i thought i might update the spec file a bit to follow work i've done myself to make everything more compliant with fedora's packaging guidelines. i've already tested these on copr and they work great!

some notable highlights include:

- patches for cmark no longer being required
  - for context, [#2166815](https://bugzilla.redhat.com/show_bug.cgi?id=2166815) on redhat's bugzilla was closed by introducing a patch that made my workaround obsolete. i kept it for a bit as the new release of cmark hadn't reached f37 and was not going to reach f36 (which was still supported at the time). i probably should've remove this during the f38 branching, but hey better late than never :)

- the jdk is now set to version 17 explicitly
  - jdk > 17 hasn't hit fedora's repos yet, but i thought this would be a good idea for the future as building with versions >17 is completely unsupported by upstream and does not work in many cases
 
- most pre/post scriptlets will only run on rhel versions < 9
  - during rhel 8, red hat was working to move most of the manual scriptlets we use here, but were not completed. with rhel 9 though, these scripts will automatically be run, same as versions of fedora > 35

- tests are now enable
  - i have no clue if this will actually work. we can only hope :)


lleyton (approved):
LGTM

PR #520, opened 2023-06-12
fix: terra-fractureiser-detector

madonuko:


anda/others/terra-fractureiser-detector/dialog.py
```diff
@@ -86,7 +86,7 @@ def on_activate(self, app):
 
 import os
 
-for home in os.listdir("/home/"):
-	try:
-		os.remove(f"/home/{home}/.config/autostart/terra-fractureiser-detector.desktop")
-	except: pass
+try:
+	home = os.environ["HOME_DIR"]
```
lleyton:
```suggestion
	home = os.path.expanduser('~')
```

This should be more reliable, I think?

lleyton (changes requested):

lleyton (approved):

PR #553, opened 2023-06-15
add: warpd

madonuko:


korewaChino (changes requested):
use %{_bindir} instead of `/usr/local/bin`

korewaChino (approved):

PR #569, opened 2023-06-16
fix(ayatana-indicator-session): include systemd session

cat-master21:
The systemd file got removed for some reason.


madonuko (approved):

madonuko:
I think I removed it by accident, my bad :3

PR #588, opened 2023-06-18
fix: uxplay

madonuko:


anda/others/uxplay/uxplay.spec
```diff
@@ -29,6 +29,7 @@ Comment=%summary
 EOF
 
 %build
+export CFLAGS="$CFLAGS -DNO_MARCH_NATIVE=ON"
 %cmake .
```
korewaChino: Do `%cmake . -DNO_MARCH_NATIVE=ON`

korewaChino (approved):

PR #597, opened 2023-06-20
bump: libappimage

madonuko:
Upstream uses the tag `1.0.4-5` and GitHub treats it as if it's earlier than `1.0.4`. Therefore, update.rhai is temporarily disabled.


nothingneko (approved):
All LGTM

PR #599, opened 2023-06-21
feature: Configure caching

korewaChino:
Configures caching for even faster builds


lleyton (changes requested):
We should probably add the package name as part of the cache key, as to prevent a single cache entry from growing too large (from being shared with all packages.)

lleyton (approved):

PR #636, opened 2023-07-19
fix: armcord

madonuko:


anda/apps/armcord/armcord.spec
```diff
@@ -10,7 +10,7 @@ Group:			Applications/Internet
 Source0:		%url/archive/refs/tags/v%version.tar.gz
 Source1:		launch.sh
 Requires:		electron xdg-utils
-BuildRequires:	nodejs-npm
```
lexisother, line 13: Formatting.

lexisother (changes requested):

anda/apps/armcord/armcord.spec
```diff
@@ -36,6 +36,7 @@ EOF
 
 
 %build
+npx pnpm@7 install --no-frozen-lockfile
```
lexisother, line 39: why are we installing deps twice?

madonuko: should be fixed

lexisother (approved):

PR #638, opened 2023-07-20
fix: graalvm

madonuko:


lleyton:
Maybe need to explicitly specify java 11/17 as a dep?

infinitebash:
`graal-jvmci-8` doesn't seem to be maintained anymore ([see here](https://github.com/graalvm/graal-jvmci-8)). It might make more sense to get a binary build from [here](https://github.com/graalvm/graalvm-ce-builds/releases).

halfcyan:
can this be closed?

Owen-sz:
Yeah probably

PR #650, opened 2023-07-24
add: pqmarble

madonuko:


anda/lib/pqmarble/pqmarble.spec
```diff
@@ -0,0 +1,52 @@
+%global commit f240b2ec7d5cdacb8fdcc553703420dc5101ffdb
+
+Name:			pqmarble
+Version:		42^%commit
```
korewaChino: make it 2.0.0 like it is supposed to be upstream???

korewaChino (approved):

PR #660, opened 2023-07-27
chore: sync with Terra 38

madonuko:


madonuko:
Bad news: `paper-plane` won't be on Terra 37 because it depends on `libadwaita-nightly` which depends on `pkgconfig(gio-2.0) >= 2.76.0` but latest on Fedora 37 is `2.74.7`.

nothingneko (approved):
Looks good, but let's try to get Paper Plane fixed if possible

PR #709, opened 2023-08-29
fix: wingpanel

madonuko:
Removes patches


lleyton (approved):
okay, yeah the check is just broke

PR #715, opened 2023-09-16
fix: actually install the kwin package (kwin-system76-scheduler-integration)

lleyton:
Everything seems to work, but I still need to figure out how to get the kwin script to be enabled by default... you'd think the kservices5 file would do it, but I guess not?


lleyton:
Update: I'm going to mark this as ready, I realize that there's no reason this package should autoenable the script, that's more in place for the user or distro.

madonuko (approved):

PR #721, opened 2023-09-26
system76-scheduler: Move default config files to /usr/share

korewaChino:


korewaChino:
Should I also modify the default Pop!_OS config preset for Fedora?

madonuko (approved):

PR #725, opened 2023-09-30
fix: ruff

madonuko:


madonuko:
I was like "where the hell are the checks?" and then "oh nvm gh is just too slow"

PR #726, opened 2023-10-07
fix: rnote

madonuko:


madonuko:
From repology, none of the distros can get 1.9.0 packaged successfully, so I figured it's an upstream issue.

PR #744, opened 2023-10-19
add: uutils-coreutils

madonuko:


korewaChino (approved):

korewaChino (dismissed):

nothingneko (dismissed):
LGTM!!

anda/tools/uutils-coreutils/uutils-coreutils.spec
```diff
@@ -0,0 +1,49 @@
+Name:			uutils-coreutils
+Version:		0.0.22
+Release:		1%?dist
+Summary:		Cross-platform Rust rewrite of the GNU coreutils
+License:		MIT
+URL:			https://github.com/uutils/coreutils
+Source0:		%url/archive/refs/tags/%version.tar.gz
+Requires:		glibc gcc-devel
```
lleyton: gcc-devel is not a real package, what are you actually trying to pull in?

lleyton (changes requested):

lleyton (changes requested):
good and bad news

good: it installs now
bad: I'm not sure why we have the uu- prefix on the binaries (and don't symlink)

korewaChino (dismissed):

korewaChino:
Many packages require coreutils itself. We need to somehow make the dep solver know that uutils can replace them all.

lleyton (dismissed):

lleyton (approved):

PR #750, opened 2023-10-23
fix and rename systemd sysusers

korewaChino:


lleyton (dismissed):

anda/others/katsu-systemd-sysusers-presets/katsu-systemd-sysusers-presets.spec
```diff
@@ -0,0 +1,72 @@
+%define debug_package %nil
+
+Name:			katsu-systemd-sysusers-presets
+Version:		1.0
+Release:		%autorelease
+Summary:		A set of systemd system user presets to supplement Fedora's defaults
```
lleyton: append ", used by Katsu" ?

lleyton (changes requested):

lleyton (approved):

PR #754, opened 2023-10-28
fix: limine

madonuko:
rename LICENSE to COPYING

### DIFF
--- a/anda/tools/limine/limine.spec
+++ b/anda/tools/limine/limine.spec
@@ -22,7 +22,7 @@ the reference implementation for the Limine boot protocol.
 %make_install
 
 %files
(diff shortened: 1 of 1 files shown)


anda/tools/limine/limine.spec
```diff
@@ -22,8 +22,8 @@ the reference implementation for the Limine boot protocol.
 %make_install
 
 %files
-%doc README.md CONFIG.md PHILOSOPHY.md PROTOCOL.md LICENSE
-%license LICENSE
+%doc README.md CONFIG.md PHILOSOPHY.md PROTOCOL.md COPYING
+%license COPYING
```
korewaChino, line 26: but the build cant find the file here

korewaChino (approved):

PR #755, opened 2023-10-28
fix(libunity-misc): make a link from libunity for unity-shell

cat-master21:
This adds a link for `libunity-protocol-private.so.0.0.0` in [libunity](https://packages.fedoraproject.org/pkgs/libunity/libunity/) to fix `unity-shell` not building. This could not be done in `unity-shell` itself because the file needs to be in `/usr/lib64/libunity-protocol-private.so` during build time which would mean `sudo`/root would have to be used and the file is out of buildroot.
**After Merged,** Please schedule a rebuild of `unity-shell`.


korewaChino (approved):

korewaChino:
lgtm, will rebuild

PR #764, opened 2023-10-29
fix: anda-srpm-macros

madonuko:


madonuko:
will need to rebuild all rust packages or sth

korewaChino (approved):

PR #782, opened 2023-11-06
add: nvidia-patch

### LINKED ISSUE #601: [Request] nvidia-patch
NVIDIA driver patch https://github.com/keylase/nvidia-patch

**Describe the motivation**
A metapackage that will have a posttrans/postinst hook for `akmod-nvidia` to patch the drivers on update time

**Additional context**
Might need some scuffed hooks

korewaChino:
I honestly don't know if we're legally allowed to add this or not, I'd just assume it's in the same case as the NVIDIA drivers.

### DIFF
--- a/anda/others/nvidia-patch/anda.hcl
+++ b/anda/others/nvidia-patch/anda.hcl
@@ -1,0 +1,6 @@
+project "pkg" {
+    rpm {
+        spec = "nvidia-patch.spec"
+        nightly = "1"
+    }
+}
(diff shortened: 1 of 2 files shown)


lainsce:
I'd say this is something that shouldn't be added as it's a custom patch not approved by NVidia

madonuko:
but what would Nvidia sue us for?

nothingneko:
We could just do it until they cease+desist us

lleyton (approved):

PR #821, opened 2023-11-20
fix: kwin-system76-scheduler-integration

madonuko:
Added script to enable the kwin script automatically

### DIFF
--- a/anda/others/kwin-system76-scheduler-integration/enable-kwin-system76-scheduler-integration.service
+++ b/anda/others/kwin-system76-scheduler-integration/enable-kwin-system76-scheduler-integration.service
@@ -3,7 +3,7 @@ Description=Enable kwin-system76-scheduler-integration
 After=network.target
 
 [Service]
-Type=simple
+Type=oneshot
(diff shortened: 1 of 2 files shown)


lleyton (changes requested):
Doesn't work, we already try something like this in Ultramarine, similarly modifying the /usr/sharekde-settings/kde-profile/default/share/config/kwinrc path, with no success.

anda/others/kwin-system76-scheduler-integration/enable-kwin-system76-scheduler-integration.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Enable kwin-system76-scheduler-integration
+After=network.target
+
+[Service]
+Type=simple
```
korewaChino: use oneshot service

korewaChino (approved):

korewaChino (dismissed):

korewaChino (approved):
hacky, but should work

korewaChino:
superceded by https://github.com/Ultramarine-Linux/ultramarine-pkgs/commit/7882ee51e67bde081eabc89075c3fc8e71817d3d

PR #826, opened 2023-11-22
fix: pantheon-session-settings

madonuko:


korewaChino (approved):
should hopefully fix the dependency hell

PR #827, opened 2023-11-22
Fix dependency issue with gala and gnome-disk-utility

korewaChino:
The previous dependency hell fix was making Pantheon sessions unbootable... After some debugging, I figured out that elementary, in their infinite wisdom, made Gala have a hard runtime dependency on... GNOME Disks. its DBus service specifically.

This finally makes Pantheon bootable again.


lleyton (approved):
this is absolutely cursed, thanks elementary

PR #828, opened 2023-11-24
fix: move system76-scheduler-dbus-proxy.sh to /usr

lleyton:


korewaChino:
actually /usr/libexec makes sense for this kind of script

lleyton:
Actually, good point

korewaChino (approved):

PR #834, opened 2023-12-02
fix: alsa-ucm-cros

madonuko:
BuildArch should be noarch


WeirdTreeThing (approved):

MeowcaTheoRange (approved):
Indeed... noarch moment

ellyq (approved):
UCMs aren't architecture-specific, so it's a good idea to build one package for all (x86_64, ARM, ARM64)

lleyton (approved):

nothingneko (commented):
lgtm

PR #863, opened 2023-12-29
fix: opentabletdriver

madonuko:


lleyton (approved):
sketchy, but makes sense

PR #864, opened 2023-12-29
fix: ruff

madonuko:


anda/langs/python/ruff/python3-ruff.spec
```diff
@@ -18,7 +18,7 @@ integrating more functionality behind a single, common interface.
 %autosetup -n ruff-%{version}
 
 %build
-maturin build --release --strip --locked --all-features
+maturin build --release --strip --all-features # --locked
```
lleyton, line 21: I don't like this, but the lock file in the release needs to be updated... so what can we really do?

lleyton (approved):

PR #866, opened 2023-12-29
fix: opentabletdriver

madonuko:
Adds post and preun scripts for enabling/disabling systemd user service


anda/system/opentabletdriver/opentabletdriver.spec
```diff
@@ -43,6 +43,11 @@ rm -rf ./dist
 mkdir -p "%{buildroot}/%{_prefix}/lib/"
 cp -r bin "%{buildroot}/%{_prefix}/lib/opentabletdriver"
 
+%post
+%systemd_user_post opentabletdriver.service
```
lleyton: %{name}.service mebbe

lleyton (changes requested):

lleyton:
image
strange...

anda/system/opentabletdriver/opentabletdriver.spec
```diff
@@ -43,6 +44,11 @@ rm -rf ./dist
 mkdir -p "%{buildroot}/%{_prefix}/lib/"
 cp -r bin "%{buildroot}/%{_prefix}/lib/opentabletdriver"
 
+%post
+%systemd_user_post %name.service
+
+%preun
+%systemd_user_preun %name.service
```
lleyton, line 49: why is there no postun?

lleyton (changes requested):
I don't even know how, but the same issue occurs

lleyton (approved):

PR #871, opened 2023-12-31
fix: nushell

infinitebash:
Remove non-existent dependency and correct post-install script. Tested and installed on Fedora 39.


nothingneko (approved):
lgtm!!

PR #911, opened 2024-02-06
fix: move submarine kpart away from /boot

lleyton:
there's no need to have this in boot, an external program will write the bootloader kpart onto a parition


madonuko:
Can you also fix the update script to use gh_tag() instead?

madonuko (approved):

PR #952, opened 2024-02-19
fix: switchboard-plug-wallet

madonuko:


lleyton:
this package is deprecated upstream, we should obsolete

madonuko:
obsoleting what?

PR #966, opened 2024-03-03
add: prime-run

korewaChino:
Simple bash script to force PRIME offloading. QoL for Arch Linux users moving to Fedora

### DIFF
--- a/anda/misc/prime-run/anda.hcl
+++ b/anda/misc/prime-run/anda.hcl
@@ -1,4 +1,5 @@
 project pkg {
+    arches = ["x86_64"]
 	rpm {
 		spec = "prime-run.spec"
 	}
--- a/anda/misc/prime-run/anda.hcl
+++ b/anda/misc/prime-run/anda.hcl
@@ -1,0 +1,5 @@
+project pkg {
+	rpm {
+		spec = "prime-run.spec"
+	}
+}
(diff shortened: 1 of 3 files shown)


anda/misc/prime-run/anda.hcl
```diff
@@ -0,0 +1,5 @@
+project pkg {
+	rpm {
```
madonuko, line 3:
```suggestion
    arches = ["x86_64"]
	rpm {
```

anda/misc/prime-run/prime-run.spec
```diff
@@ -0,0 +1,38 @@
+Name:           prime-run
+Version:        0.0.1
```
madonuko:
```suggestion
Version:        1.0
```
Probably want to follow other distros' versioning

anda/misc/prime-run/prime-run.spec
```diff
@@ -0,0 +1,38 @@
...
+Summary:        A simple script to run an application with NVIDIA PRIME GPU offloading
+
+License:        MIT
+URL:            https://terra.fyralabs.com
+Source0:        prime-run.sh
+
+# Bash script
+Requires:       bash
+BuildArch:      noarch
+Supplements:    nvidia-driver
```
madonuko: `Supplements:` only work if `nvidia-driver` requires this package; in this case there's probably no effect

madonuko (approved):

PR #969, opened 2024-03-05
feat: package budgie-indicator-applet

lleyton:


lleyton:
doesn't even work lmao

PR #990, opened 2024-03-13
fix(vala-panel-appmenu): include jayatana files

cat-master21:
Jayatana is an actual project by Google https://code.google.com/archive/p/java-swing-ayatana/ but it should be fine.


madonuko (dismissed):

madonuko (approved):

madonuko:
tysm

PR #1000, opened 2024-03-25
chore: obsolete sysusers misc packages

lleyton:
these packages solved an issue upstream, which has been fixed in 40 and rawhide, as they're shipping their own sysusers presets now


madonuko (approved):
Happy 1000!

PR #1036, opened 2024-04-02
Port various f39 fixes to rawhide

lleyton:
Am I missing anything here? (besides things that autoupdate will handle, once we do the branch switch)


madonuko (approved):

korewaChino (approved):
ok lgtm

PR #1037, opened 2024-04-02
Add package for fast-syntax-highlighting for ZSH

VehementHam:


anda/tools/F-Sy-H/anda.hcl
```diff
@@ -0,0 +1,5 @@
+project pkg {
```
madonuko:
```suggestion
project pkg {
  arches = ["x86_64"]
```

anda/tools/F-Sy-H/F-Sy-H.spec
```diff
@@ -0,0 +1,31 @@
...
+Release: 1%?dist
+
+Summary: Feature-rich Syntax Highlighting for Zsh
+License: BSD-3-Clause
+URL:     https://github.com/z-shell/%name
+Source0: %url/archive/refs/tags/v%{version}.tar.gz
+
+BuildArch: noarch
+
+BuildRequires: zsh
```
lleyton: You can delete this, there isn't a build script anyways :p

VehementHam: I think those are supposed to be deletions

anda/tools/F-Sy-H/anda.hcl
madonuko: ?

madonuko: ah in case of the confusion, since this is a noarch package, we can just build this with the x86_64 builders. if we use both builders there will be duplications

VehementHam: Sorry I thought I commited it.

madonuko:
I think you applied the suggestion twice…

lleyton (changes requested):
image

Also, maybe we should require zsh?

VehementHam:
> image
> 
> Also, maybe we should require zsh?

Not sure. It was in the `.spec` originally, but I think it was removed by the recommendation of another member.

VehementHam:
Feel free to make a commit.

VehementHam:
> image
> 
> Also, maybe we should require zsh?

How are you testing this?

lleyton:
> > image
> > Also, maybe we should require zsh?
> 
> How are you testing this?

Under the checks tab, GHA artifacts containing the RPMs are uploaded there

VehementHam:
Okay, well I did not fix to `fast-syntax-highlighting` in every file yet, so that is a problem.

VehementHam:
> Okay, well I did not fix to `fast-syntax-highlighting` in every file yet, so that is a problem.

Fixed that issue

anda/tools/fast-syntax-highlighting/fast-syntax-highlighting.spec
```diff
@@ -0,0 +1,27 @@
+Name: fast-syntax-highlighting
+Version: 1.67
```
lleyton: 1.67 doesn't seem to be a valid version, which is causing the build to fail

madonuko: Yeah the latest is 1.55 I think

you'll also need to use `gh_tag` instead in `update.rhai`

VehementHam: Think I corrected the version

VehementHam: > you'll also need to use `gh_tag` instead in `update.rhai`

Think I did this to

lleyton:
getting errors when loading it, I think you're forgetting to include some files

Owen-sz:
Added in https://github.com/terrapkg/packages/pull/6712

PR #1094, opened 2024-04-15
add: sunshine

### LINKED ISSUE #1093: [New] Sunshine
https://github.com/LizardByte/Sunshine

**Describe the motivation**
While this app already has a Flatpak, direct KMS framebuffer streaming does not really work unless you set up some special process capabilities to allow it to talk to the DRM in userspace.

**Additional context**
upstream RPM is packaged using CPack

korewaChino:
The name of the software and link to website (if it exists) and source code (if available).

**Describe the motivation**
A clear and concise description of why you want this package. Would this provide value to other users?

**Additional context**
Add any other context about the package submission here. Link to any relavent issues.

### DIFF
--- a/anda/misc/sunshine/anda.hcl
+++ b/anda/misc/sunshine/anda.hcl
@@ -1,0 +1,5 @@
+project pkg {
+	rpm {
+		spec = "sunshine.spec"
+	}
+}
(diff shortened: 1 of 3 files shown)


sadlerm4:
what does it mean that ``nvidia-driver`` is a blocker?

PR #1301, opened 2024-05-30
fix(terra-release): disable repodata check to make mirroring work

madonuko:
This is done because not even Fedora enable this! This will cause issues when Terra mirrors are not syncing in real time, causing dnf to think Terra is broken when hashes don't match with tetsudou.

### DIFF
--- a/anda/terra/release/terra-release.spec
+++ b/anda/terra/release/terra-release.spec
@@ -1,6 +1,6 @@
 Name:           terra-release
 Version:        40
-Release:        3
+Release:        4
 Summary:        Release package for Terra
 
 License:        MIT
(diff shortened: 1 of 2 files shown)


lleyton:
Please don’t merge this yet, want to test some stuff. Albeit, isn’t the repomd verified using the metalink hash? Unsure if switching this to 0 changes that

korewaChino:
> Please don’t merge this yet, want to test some stuff. Albeit, isn’t the repomd verified using the metalink hash? Unsure if switching this to 0 changes that

iirc the repomd is only used from upstream, they only mirror the big RPM files so if that fails it will just try another mirror

lleyton:
If that's the case, I don't see how this change is necessary

madonuko:
> If that's the case, I don't see how this change is necessary 

No, that's exactly why this change is necessary.

Without this, dnf would only choose repos.fyralabs.com because that's the only mirror deemed to have "the correct hash" by dnf.

PR #1332, opened 2024-06-08
fix: authy

madonuko:


nothingneko (commented):
sussy, why

madonuko:
upstream no longer exists

madonuko:
+ more reproducible builds

nothingneko:
looping in @lleyton

madonuko:
Think this counts as "shipping unmaintained software, so instead we should remove this package entirely

PR #1396, opened 2024-06-27
add: python3-pillow-heif

runlevel5:
## Motivation

This package is widely used in many Python-based projects and few friends have requested me to package it for them.

## Notes

I've successfully got it built with koji but unsure how it would be with anda system. Please kindly provide me guidance if it does not get built successfully

### DIFF
--- a/anda/langs/python/pillow-heif/0001-Use-env-var-PILLOW_HEIF_VERSION-to-get-version.patch
+++ b/anda/langs/python/pillow-heif/0001-Use-env-var-PILLOW_HEIF_VERSION-to-get-version.patch
@@ -27,7 +27,7 @@ index b3bf0ce..e5365b0 100644
 -    version_file = "pillow_heif/_version.py"
 -    exec(compile(Path(version_file).read_text(encoding="utf-8"), version_file, "exec"))  # pylint: disable=exec-used
 -    return locals()["__version__"]
-+    os.environ("PILLOW_HEIF_VERSION")
++    os.environ["PILLOW_HEIF_VERSION"]
  
  
  def _cmd_exists(cmd: str) -> bool:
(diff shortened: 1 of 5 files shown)


madonuko:
I think you need `python3-rpm-macros` as a BuildDependency

runlevel5:
> I think you need python3-rpm-macros as a BuildDependency

I've just made another commit. Could you please trigger a new build for me? if it works, I will merge 2 commits into 1 and force push

runlevel5:
@madonuko to save you time from helping, could you please point me to a doc where I could actually run anda locally to make sure everything works?

madonuko:
If you have Terra installed, you could install anda via dnf.

https://developer.fyralabs.com/andaman/installation

You will also need to install terra-mock-configs

To manually run anda:
```sh
anda build anda/langs/python/python-heif/pkg -c terra-40-x86_64
```
Make sure your user is in the mock group. (`sudo usermod -a -G mock username`)

runlevel5:
@madonuko it can be built successfully with anda locally. Please help me trigger the CI build

runlevel5:
@madonuko so the issue of the build is related to Python 3.13 (which would be defaulted to on F41). I think for the time-being the package could be built against f40 and f39 only. Possible?

Lastly, how to target f41 build with anda on local machine? Many thanks

madonuko:
> I think for the time-being the package could be built against f40 and f39 only. Possible?

Yes, you can just make a new PR to terrapkg:f40 instead.

> so the issue of the build is related to Python 3.13

I'm not so sure if that's actually true…? I'd try generating this from `pyp2rpm` instead to be honest.

> how to target f41 build with anda on local machine?

We don't have mock files for f41 yet because it doesn't exist yet; Fedora 41 branches off from rawhide in August. You could target rawhide, you just need a container unfortunately.

runlevel5:
@madonuko let's have it for f40 for the time-being, I will inform the author to look into address the issue upstream and hopefully 0.17.0 would be out around the same time of f41

runlevel5:
Once the python 3.13 compatibility is addressed upstream, I will make a new PR

PR #1465, opened 2024-07-10
fix zed-nightly auto updates

june-fish:


anda/devs/zed/nightly/update.rhai
```diff
@@ -1,5 +1,6 @@
 if filters.contains("nightly") {
   rpm.global("commit", gh_commit("zed-industries/zed"));
+  rpm.global("commit_date", date());
   if rpm.changed() {
     let v = find("(\\d+\\.\\d+\\d+\\.\\d+)", find("\nversion = \"(\\d+\\.\\d+\\d+\\.\\d+)\"\n", gh_rawfile("zed-industries/zed", "main", "crates/zed/Cargo.toml"), 1), 1);
     rpm.global("ver", v);
```
lleyton, line 6: it seems called in a few lines later?

lleyton (changes requested):

anda/devs/zed/nightly/update.rhai
madonuko: This change will cause a daily rebuild… Probably not the intended change

PR #1473, opened 2024-07-10
add: komikku

runlevel5:
## What?

* Add komikku app

### DIFF
--- a/anda/apps/komikku/update.rhai
+++ b/anda/apps/komikku/update.rhai
@@ -1,0 +1,3 @@
+let latest_tag = get("https://codeberg.org/api/v1/repos/valos/Komikku/tags").json_arr()[0].name;
+let new_version = find("([\\.\\d]+)", latest_tag, 1);
+rpm.version(new_version);
--- a/anda/apps/komikku/anda.hcl
+++ b/anda/apps/komikku/anda.hcl
@@ -1,0 +1,5 @@
+project pkg {
+  rpm {
+	  spec = "komikku.spec"
+  }
+}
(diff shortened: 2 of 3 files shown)


anda/apps/komikku/update.rhai
```diff
@@ -0,0 +1 @@
+rpm.version(codeberg("valos/Komikku"));
```
runlevel5: @madonuko wondering if you could point me to where the `gh` script is so I could try to figure out if I could add a new rhai script to find out new release version of codeberg (NOTE: codeberg is yet another GitHub-like software)

madonuko: ooh, interesting…

So basically all those Rhai functions are coded in rust [in the anda repository](https://github.com/FyraLabs/anda/blob/main/andax/src/fns/tsunagu.rs)

I think instead of making a new function you could just use `get()` to access an API endpoint to fetch the version.

madonuko: List of functions baked into andax: https://developer.fyralabs.com/terra/autoupdate

if you want to experiment with andax/rhai scripts you can run an rhai file using `anda run test.rhai`; the catch is there is no `rpm` interface, so instead of `rpm.version()` you will need to `print()` things out.

madonuko:
```ts
print(get("https://codeberg.org/api/v1/repos/valos/Komikku/tags").json_arr()[0].name);
```
Is this probably what you want? Got bored so I did some API docs reading (remember to change to `rpm.version()` or something)

runlevel5:
@madonuko it is ready to go

madonuko (approved):

PR #1931, opened 2024-08-16
fix: terra-release

madonuko:


madonuko:
Do we need to sync this?

lleyton:
> Do we need to sync this?

I don't think so

lleyton (approved):

PR #777, opened 2023-11-05
add: ffmpeg

madonuko:

PR #2225, opened 2024-10-10
add: ffmpeg

madonuko:


madonuko:
also, we kinda added ffmpeg back then in #777 but not sure where it went.

nothingneko (approved):

PR #2232, opened 2024-10-10
add: x265

madonuko:


nothingneko:
is this patented?

korewaChino:
wikipedia reference: https://en.wikipedia.org/wiki/X265

nothingneko (approved):

PR #2376, opened 2024-10-24
bump: kde-liquidshell

madonuko:


sadlerm4:
i'll see if i can help? liquidshell was next on my list of packages that need fixing

madonuko:
yeah so apparently `%cmake` failed because it can't find qt…?
was confusion

PR #2523, opened 2024-11-28
fix: keyd-2.4.3

sadlerm4:
Increment release to one higher than previous highest release for version 2.4.3

### DIFF
--- a/anda/tools/keyd/keyd.spec
+++ b/anda/tools/keyd/keyd.spec
@@ -1,6 +1,6 @@
 Name:			keyd
 Version:		2.4.3
-Release:		1%?dist
+Release:		4%?dist
 Summary:		Key remapping daemon for linux
 URL:			https://github.com/rvaiya/keyd
 License:		MIT


madonuko:
conflict?

sadlerm4:
yeah, do you understand what i'm trying to do here?

sadlerm4:
when rabo bumps version back to 2.4.3 release stays at 1 so you've been building 2.4.3-1 (which isn't really 2.4.3-1, 2.4.3-1 is the initial release for terra [#459]) for some time lol

there are merge conflicts because you merged my other PR for keyd-2.5.0 - i don't see the problem you've been building both versions of keyd from the same spec (intentional or otherwise) already

i thought it only affected f41 but it affects f40 too, if you could backport too that would be amazing

this PR is mainly just a OCD thing, if you want to keep 2.4.3 in terra you might as well do it right

nothingneko (approved):

PR #2532, opened 2024-11-28
prismlauncher{,-nightly}: use built-in java manager on f42 and above

getchoo:
See commit message for context

This will need the [Adoptium repository](https://adoptium.net/installation/linux/#_centosrhelfedora_instructions) in Anda, or some other source for JDKs < 17[^1]

[^1]: An alternative here could also be dropping support for Java < 8 in compiled bytecode, but that would cause older versions of the game to break.


madonuko:
has been applied and merged over at 2467 thank you!

PR #2538, opened 2024-11-29
fix: pqmarble

sadlerm4:
upstream is no longer in active development


madonuko:
if you remove nightly it'll pull updates more frequently

nightly updates every day, normal updates every 10 minutes

sadlerm4:
oh lol that's kinda funny:D

PR #2549, opened 2024-12-03
Conflict with upstream libplacebo

korewaChino:
Conflicts terra-libplacebo with upstream version

why was this package added in the first place when https://packages.fedoraproject.org/pkgs/libplacebo/libplacebo/?


korewaChino (commented):
This PR should be updated to full removal if we are removing it

madonuko (approved):

sadlerm4:
removed in #3126

PR #2643, opened 2024-12-15
fix(mock-configs): include mock configs for el10

madonuko:


nothingneko (approved):

madonuko:
the mock files are still broken for reasons so don't merge yet

PR #2653, opened 2024-12-16
add: tmon

sadlerm4:
a cute and functional system monitor i found on r/unixporn

and yes i am aware i'm kinda using terra to practice how to package rpms 😭

### DIFF
--- a/anda/devs/tmon/tmon.spec
+++ b/anda/devs/tmon/tmon.spec
@@ -15,7 +15,7 @@ Source0:        %{url}/archive/%{commit}.tar.gz
 Requires:       lm_sensors
 Conflicts:      kernel-tools
 
-Suggests:       nerd-fonts
+Recommends:     google-noto-color-emoji-fonts
 
 BuildRequires:  make gcc-c++ ncurses-devel
 
--- a/anda/devs/tmon/tmon.spec
+++ b/anda/devs/tmon/tmon.spec
@@ -1,0 +1,46 @@
+%global commit e8cae0d88dc8d51cb0e34e3fd3553bcbdaf04ca5
+%global shortcommit %(c=%{commit}; echo ${c:0:7})
+%global commit_date 20241215
+%global debug_package %{nil}
+
+Name:           tmon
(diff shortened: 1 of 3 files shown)


anda/devs/tmon/tmon.spec
```diff
@@ -0,0 +1,46 @@
...
+%global shortcommit %(c=%{commit}; echo ${c:0:7})
+%global commit_date 20241215
+%global debug_package %{nil}
+
+Name:           tmon
+Version:        %{commit_date}.git~%{shortcommit}
+Release:        1%{?dist}
+Summary:        A tiny system monitor for Linux
+
+License:        GPL-3.0
```
madonuko:
```suggestion
License:        GPL-3.0-only
```

anda/devs/tmon/tmon.spec
```diff
@@ -0,0 +1,46 @@
...
+URL:            https://github.com/pondda/tmon
+Source0:        %{url}/archive/%{commit}.tar.gz
+
+Requires:       lm_sensors
+Conflicts:      kernel-tools
+
+Suggests:       nerd-fonts
```
madonuko: `Suggests` is useless here I think

madonuko:
also you forgot about Packager:

sadlerm4:
> also you forgot about Packager:

oh we're still doing that

anda/devs/tmon/tmon.spec
sadlerm4: Can't really tell if it needs an emoji font or nerd font installed tbh

sadlerm4: It will work without either because it can be used in a tty, but I would prefer to suggest/recommend one of the two, or perhaps even both? What do you think?

madonuko: `Suggests:` will have no effect on this
`Recommends:` will pull in all `nerd-fonts`
soooooo *dunno*

sadlerm4: ah i see what you mean now

madonuko (approved):

PR #2758, opened 2024-12-26
fix(anki): enable aarch64 support

madonuko:


madonuko:
https://github.com/terrapkg/packages/actions/runs/12562995462/job/35024241115?pr=2758#step:7:21908
doesn't make sense?

PR #2794, opened 2024-12-30
Fix: telescope noarch build

Owen-sz:
Make telescope build on arm


madonuko:
this package is `BuildArch: noarch`. a `.noarch` package can be built once on x86_64, and installed on all architectures.

Owen-sz:
> this package is `BuildArch: noarch`. a `.noarch` package can be built once on x86_64, and installed on all architectures.

Ohh, good to know

PR #2878, opened 2025-01-04
mesa: Obsolete freeworld drivers

korewaChino:
Allows people who were already using `mesa-freeworld` to upgrade to our Mesa builds


korewaChino:
merged into #2879

Issue #2877, opened 2025-01-04
Repackage negativo17 packages


korewaChino:
Reference: https://negativo17.org/repos/fedora-multimedia.repo

- [ ] apps/blender-{cuda,hip}
- [ ] tools/CDRtools
- [ ] system/intel-vaapi-driver
- [ ] system/intel-mediasdk
- [ ] system/intel-gmmlib
- [ ] system/akmod-evdi
- [ ] system/akmod-nullfs
- [x] system/akmod-xone
- [ ] tools/DirectXShaderCompiler
- [ ] system/displaylink
- [x] ~~lib/libav*~~ ffmpeg
- [x] lib/gstreamer-plugins-bad
- [x] lib/gstreamer/plugins-ugly
- [ ] apps/vlc (Upstream VLC is stripped down)

korewaChino:
not yet done

madonuko:
ffmpeg is packaged tho?

PR #2888, opened 2025-01-05
Add: GStreamer Plugins

GildedRoach:
Related issue: #2877


madonuko:
https://github.com/terrapkg/packages/blob/frawhide/anda/desktops/compiz9/update.rhai

maybe check this out?

GildedRoach:
> https://github.com/terrapkg/packages/blob/frawhide/anda/desktops/compiz9/update.rhai
> 
> maybe check this out?

Trying to figure out how to do this, struggling a tad on the `find` syntax but working on it.

Also so annoying one of the packages here requires the `-devel` of another here and another requires another dep I am yet to even package so far (`ipu6-kmod`). I'll work on fixing the latter in a second but the CI will still fail due to not finding the build deps.

madonuko:
image
```ts
let html = get("https://gstreamer.freedesktop.org/src/gst-libav/?C=M;O=D");
print(find(`gst-libav-([\d.]+)\.tar\.xz`, html, 1));
```

GildedRoach:
Tried to make an `if` statement based on version for when Rawhide inevitably has a newer version than other branches but it's not working out and keeps breaking Meson so I'll just babysit this package I guess.

GildedRoach:
At long last the update path is secured for the GStreamer 1.15.x transition whenever that happens, if everything else is fine I think this thing is finally ready to merge. :sob: 

My God I hate the `%if` macro.

madonuko (approved):

madonuko:
get yourself the sync labels if necessary

GildedRoach:
Thank God for squash and merge. 💀

PR #3049, opened 2025-01-18
Change: Rename Gamescope to terra-gamescope, move to main repo now that the name no longer conflicts

GildedRoach:


madonuko:
maybe have a sub package called `gamescope` that requires terra-extras and terra-gamescope?

GildedRoach:
> maybe have a sub package called `gamescope` that requires terra-extras and terra-gamescope?

Can look into doing this tomorrow.

PR #3026, opened 2025-01-16
Rebase FFmpeg to Negativo17's version'

korewaChino:

PR #3112, opened 2025-01-24
Add: xavs2

GildedRoach:
Needed for #3026


korewaChino:
let's not do the update incident again, rawhide only plox

korewaChino (approved):

GildedRoach:
> let's not do the update incident again, rawhide only plox

Oh RPM Fusion doesn't carry this package and `xavs` not 2 is in 41 so that's why. Either is fine.

PR #3138, opened 2025-01-27
chore(f42): add packages to be obsoleted

sadlerm4:
merge when 42 is branched

### DIFF
--- a/anda/terra/obsolete/terra-obsolete.spec
+++ b/anda/terra/obsolete/terra-obsolete.spec
@@ -121,15 +121,31 @@ BuildArch:  noarch
 %obsolete_ticket https://github.com/terrapkg/packages/issues/991
 %obsolete iosevka-fusion-fonts 25.1.1-2
 
-# obsolete_ticket <github PR URL>
-# %obsolete terra-sddm 0.21.0-6
-# %obsolete terra-rgbds
-# %obsolete terra-libplacebo
-# %obsolete terra-wl-clipboard
-# %obsolete terra-maturin
-# %obsolete terra-libindicator
-# %obsolete terra-blueprint-compiler
-# %obsolete nushell
+%obsolete_ticket https://github.com/terrapkg/packages/pull/3138
+%obsolete terra-sddm 0.21.0-6
+%obsolete terra-rgbds 0.9.0-3
+%obsolete terra-libplacebo 7.349.0-3
+%obsolete terra-libplacebo-devel 7.349.0-3
+%obsolete terra-wl-clipboard 2.2.1-2
+%obsolete terra-maturin 1.8.1-3
+%obsolete terra-libindicator 16.10.0-3
+%obsolete terra-libindicator-devel 16.10.0-3
+%obsolete terra-libindicator-gtk3 16.10.0-3
+%obsolete terra-libindicator-gtk3-devel 16.10.0-3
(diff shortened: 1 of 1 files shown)


madonuko:
this does not make sense, if these packages are not in f42, then there's nothing to obsolete

sadlerm4:
pantheon packages that need to be obsoleted too:
- elementary-appcenter
- switchboard-plug-datetime
- switchboard-plug-locale
- switchboard-plug-parental-controls
- switchboard-plug-power
- switchboard-plug-security-privacy
- switchboard-plug-useraccounts
- switchboard-plug-wacom

sadlerm4:
> this does not make sense, if these packages are not in f42, then there's nothing to obsolete

what about upgrades

sadlerm4:
i'm not entirely sure what happens during system-upgrade when a package doesn't exist in the new releasever

sadlerm4:
> ### fedora-obsolete-packages
> 
> Consider what will happen to systems which have the now-retired packages installed. Generally, such packages will simply remain on the system as it is updated, becoming increasingly outdated. If this is not acceptable, such as if there are upgrade path issues or security concerns, add your package to [fedora-obsolete-packages](https://src.fedoraproject.org/rpms/fedora-obsolete-packages/).

literally all of these terra-* packages potentially represent upgrade path issues because they all conflict with upstream, and you can't guarantee that zero users have them installed, even if the number is probably <1%

how is it even possible that terra-sddm was approved without conflicting/providing sddm?????????

lea was right this has long-term consequences for um42 and beyond until you can actually guarantee that none of these packages are installed on users' systems anymore so the best way (that i can see anyway) is to make sure they get removed

sadlerm4:
sorry i'm being overly harsh it's just i didn't expect to still be finding packages that are conflicting with upstream like a week later

> i'm not entirely sure what happens during system-upgrade when a package doesn't exist in the new releasever

again if packages get removed during system-upgrade then you're right ofc it's a non-issue then and i'll close this PR

GildedRoach:
> sorry i'm being overly harsh it's just i didn't expect to still be finding packages that are conflicting with upstream like a week later
> 
> > i'm not entirely sure what happens during system-upgrade when a package doesn't exist in the new releasever
> 
> again if packages get removed during system-upgrade then you're right ofc it's a non-issue then and i'll close this PR

The biggest problem is when packages were dependencies of something else that is trying to update, so you are right it _can_ cause issues.

madonuko:
> what about upgrades

they are not removed, you are breaking installations if you remove them

sadlerm4:
> > what about upgrades
> 
> they are not removed, you are breaking installations if you remove them

thank you, hence the point of obsoleting

i do not want new users installing these packages (hence satm rpm rm) and i want 40/41 users who still have these packages installed to have them removed when upgrading to 42

frankly i would do obsoleting in 41 but that is frowned upon since it's already released

if this is somehow unacceptable, then you should really let me know ASAP because then we should be talking about adding conflicts in 41 to all of these terra-* packages (which i didn't think to do because most of the specs had been removed already)

madonuko:
I see then; you have my ack + approval for the obsolete

sadlerm4:
will revisit in 5 months /j

sadlerm4:
- uutils-coreutils-util-linux
- uutils-coreutils-util-linux-replace

madonuko (approved):

PR #3122, opened 2025-01-25
Fix (Ghostty): New build flags for `.terminfo` and `.termcap`

GildedRoach:

PR #3161, opened 2025-01-29
[f40] Fix (Ghostty): New build flags for .terminfo and .termcap (#3122)

GildedRoach:
Nya nya meow and whatever


GildedRoach:
Oh right Zig just dies on F40 still :+1:

GildedRoach:
> Oh right Zig just dies on F40 still 👍

Turns out it was the RPM version.

madonuko (approved):

PR #3158, opened 2025-01-28
fix(maple-fonts): rpm < 4.20 limitation

sadlerm4:
doing it this way since i don't want the spec to be different between branches

PR #3168, opened 2025-01-30
fix(maple-fonts): fix string manipulation of upstream version string

sadlerm4:
i guess i should thank rabo for forcing me to learn ``sed`` ffs


sadlerm4:
will backport manually together with #3158

madonuko (approved):

PR #3205, opened 2025-02-05
chore(devcontainer): simplify image

GarciaLnk:
I was having issues running `anda build` with both mock and rpmbuild through the devcontainer. This simplifies the devcontainer using the builder image as the base and adding the common-utils feature (which takes care of creating the vscode user and more).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,9 +1,0 @@
-ARG VARIANT="rawhide"
-FROM fedora:${VARIANT}
-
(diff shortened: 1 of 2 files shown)


madonuko:
why did you remove the dockerfile?

GarciaLnk:
It's not used anymore, `ghcr.io/terrapkg/builder:frawhide` is used instead and set directly in the `devcontainer.json`

madonuko (approved):

PR #3275, opened 2025-02-11
Fix (scenefx): Add missing build dependencies

GildedRoach:
Do not merge until #3274 is merged.

`cmake` added to the build shuts up about errors, not actually used.

PR #3280, opened 2025-02-11
[el10] Fix (scenefx): Add missing build dependencies (#3275)

GildedRoach:
Meow.


GildedRoach:
Well that's not good.

PR #3419, opened 2025-02-17
fix(carla): unnecessary builddepends

sadlerm4:


sadlerm4:
okay i'll try building locally later

error is very illuminating though, hopefully it builds successfully with python3-file-magic instead of python3-magic, because python3-file-magic seems to already be installed on the docker image you're using, hence the conflict

madonuko:
you are looking for BuildConflicts

PR #3549, opened 2025-02-25
fix(gradle): use java-latest-openjdk

madonuko:


GildedRoach:
The URLs cannot be the version because the links do not end in `.0`, maybe a global macro to remove the last decimal?

madonuko:
should use adoptium instead if possible

metcya:
superseded by #8012

PR #3589, opened 2025-02-28
docs: atomic distro install guides

boredsquirrel:
run0 is a bit broken, so pipes don't work (I think?)
but it works without setuid which is nice

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,10 @@ sudo dnf install --nogpgcheck --repofrompath 'terra,https://repos.fyralabs.com/t
 If you are using immutable/atomic editions of Fedora, run the following commands instead:
 
 ```bash
-curl -fsSL https://github.com/terrapkg/subatomic-repos/raw/main/terra.repo | pkexec tee /etc/yum.repos.d/terra.repo
-sudo rpm-ostree install terra-release
+curl -fsSLO https://github.com/terrapkg/subatomic-repos/raw/main/terra.repo
+run0 mv terra.repo /etc/yum.repos.d/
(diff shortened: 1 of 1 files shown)


nothingneko:
why run0 instead of sudo or pkexec?

boredsquirrel:
run0 works without suid. pkexec and sudo both need that. So on systems where
- A suid binaries are removed
- B the main user is not in wheel

this works

GildedRoach:
> run0 works without suid. pkexec and sudo both need that. So on systems where
> 
>     * A suid binaries are removed
> 
>     * B the main user is not in wheel
> 
> 
> this works

I do not know of any Fedora atomic spin that does not have `sudo` and in my personal opinion a user not being able to add repos without being in the wheel group is by design so I find that questionable.

Can you please name an example of a Fedora atomic distro for the first point?

nothingneko:
I agree, a user without explicit permission (wheel group membership) should not be able to install a repository

GildedRoach:
I would go as far as to say that is bypassing system security when Fedora's security measures are built around `sudo`. `run0` in my personal opinion should *not* be in an install guide for a system intending you to use `sudo` or `pkexec`. `run0` needs different security configuration.

Also, what about systems without `systemd`? I would say those are more common than systems without `sudo`.

nothingneko:
> Also, what about systems without `systemd`? I would say those are more common than systems without `sudo`.
Neither are particularly common in the RH ecosystem

Overall, this is a bad idea, and adds more steps to the installation guide for a feature (installing a repo without sudo) most people don't need. In addition the security implications are terrible, we will not be merging this change.

Issue #3507, opened 2025-02-22
[Request] openrazer


dreamyukii:
**What software are you requesting to be packaged?**

[openrazer](https://openrazer.github.io/) , software for managing Razer hardware and peripheral

**Describe the motivation**

upstream only support dkms for openrazer kernel module, not suitable for Fedora Atomic Desktop 

**Additional context**

maybe need some modification to support akmods

madonuko:
we'll package the dkms for now

PR #3662, opened 2025-03-02
add: openrazer

madonuko:


madonuko:
#3507 (only the dkms part)

GildedRoach:
This is...certainly a DKMS spec but if it's official I would *hope* it works.

The Source0 URL is: https://github.com/openrazer/openrazer/archive/refs/tags/v%{version}.tar.gz

You may want to split `common` and maybe the other non-DKMS subpackages into a separate spec called `openrazer` if we want to support an Akmod package later on.

It's not an official requirement but most Fedora DKMS packages prepend `dkms-` instead of suffixing it to contrast with `akmod-` though I know we aren't obligated to follow this.

dreamyukii:
for the kmod ublue already package the kernel module as [kmod module](https://github.com/ublue-os/openrazer/blob/master/openrazer-kmod-common.spec)

GildedRoach:
> for the kmod ublue already package the kernel module as [kmod module](https://github.com/ublue-os/openrazer/blob/master/openrazer-kmod-common.spec)

If we get this building I can add the kmod stuff.

PR #3773, opened 2025-03-07
Fix: stardust server

Owen-sz:
Add runtime libraries and other packages needed for misc stardust server usages

### DIFF
--- a/anda/stardust/server/stardust-server.spec
+++ b/anda/stardust/server/stardust-server.spec
@@ -6,7 +6,7 @@
 
 Name:           stardust-xr-server
 Version:        %commit_date.%shortcommit
-Release:        2%?dist
+Release:        3%?dist
 Summary:        Usable Linux display server that reinvents human-computer interaction for all kinds of XR.
 URL:            https://github.com/StardustXR/server
 Source0:        %url/archive/%commit/server-%commit.tar.gz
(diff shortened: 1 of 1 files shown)


GildedRoach:
Are all of these required for the server to function or could some be a `Suggests:` instead?

Owen-sz:
They're needed if you wanna load or do anything in the server beyond very basic functionality.

anda/stardust/server/stardust-server.spec
```diff
@@ -17,6 +17,11 @@ BuildRequires:  glx-utils fontconfig-devel glibc libxcb-devel wayland-devel
 BuildRequires:  openxr-devel libglvnd-devel libglvnd-gles mesa-libgbm-devel
 BuildRequires:  libwayland-egl libX11-devel libXfixes-devel
 BuildRequires:  mesa-libEGL-devel libxkbcommon-devel
+
+Requires:       libxkbcommon libstdc++ libX11 libXfixes libEGL libgbm fontconfig pcre2
+Requires:       libgcc glibc libxcb libglvnd libdrm expat freetype libxml2 libXau libbrotli
+Requires:       zlib-ng-compat bzip2-libs libpng harfbuzz libbrotli xz-libs glib2 graphite2
```
madonuko, line 23: maybe remove things listed in buildreqs already?

GildedRoach:
> They're needed if you wanna load or do anything in the server beyond very basic functionality.

That is fair, but generally things not needed for the program itself to run should be weak deps so people can exclude the packages they may not want for any reason (not using a specific feature, etc.)

anda/stardust/server/stardust-server.spec
Owen-sz: The binary links against all of these, so wouldn't they need to all be runtime deps?

Owen-sz:
> > They're needed if you wanna load or do anything in the server beyond very basic functionality.
> 
> That is fair, but generally things not needed for the program itself to run should be weak deps so people can exclude the packages they may not want for any reason (not using a specific feature, etc.)

Theoretically they should all be needed for full functionality.

anda/stardust/server/stardust-server.spec
madonuko: RPM detects all the shared dependencies needed, which you can check with `dnf rq --requires ./path/to/idk.rpm`

the Requires: field is specifically for things that are not linked. For example if your binary runs vim inside bash then both vim and bash should be specified as a dependency since they are not linked.

Remember last time we tested stardust with Nova? Well I suspect everything has already been specified correctly by rpm automatically since it ran "fine"

If you see something like "undefined symbol" or "symbol lookup error" then 1. there's something wrong with the linking process and/or 2. you might need to specify some extra stuff in Requires

You may choose to continue specifying all the runtime dependencies but it probably does nothing and you increase maintenance burden, since if a newer version doesn't need a dep anymore you'll need to manually remove the Requires

Owen-sz: Ahh, I see thank you. I'll do some more testing and talk to Nova.

PR #3987, opened 2025-03-18
refactor(heroic): Changes for aarch64 preparation, Git clone to install Git hooks

GildedRoach:
This spec is very cursed now. Heroic does *not* build as aarch64 yet but they are beginning to work towards it on Linux and some aarch64 binaries now build (regrettably, regardless of architecture, which confused RPM AutoReq pretty badly).

Changes:
- Make Heroic not dep on GCC or incorrect architecture deps *without* disabling AutoReq (!!!)
- Git clone Heroic so Husky Git hooks install (Heroic uses these to check for updates, as far as I can tell both for itself but also its WINE manager)
- Drop the afterPack patch since this is kinda useless when not on Mac anyway
- Install (almost, there's a ton of duplicate, exactly the same ones for Node modules but not Heroic or its Electron runtime I excluded) *all* licenses for Heroic in the license dir
- Remove the incorrect architecture binaries
- Removed ExclusiveArch in preparation for whenever Heroic gets an aarch64 release (anda.hcl remains the same)
- Add bundled binaries with versions (this sort of matters mostly because Fedora packages Legendary)
- Removed the `extest` weak dep for now since while I've heard when built as x86_64 it helps with launchers like Heroic I can't confirm for sure at this time (needs GNOME testing maybe?)
- Fixed the changelog order blunder
- Switch from Pacman target to generic Linux target, this builds for both AppImage *and* Snap right now so the unpacked binary *shouldn't* attempt to self update and it requires less build deps and build time
- Added a symlink to make `heroic` executable mainly because of what the actual binary is called

### DIFF
--- a/anda/games/heroic-games-launcher/heroic-games-launcher.spec
+++ b/anda/games/heroic-games-launcher/heroic-games-launcher.spec
@@ -51,7 +51,6 @@ Packager:      Gilver E. <rockgrub@disroot.org>
 Heroic is a Free and Open Source Epic, GOG, and Amazon Prime Games launcher for Linux, Windows, and macOS.
 
 %prep
-rm -rf ./*
 %git_clone https://github.com/%{org_name}/%{git_name} v%{version}
 sed -i 's/Exec=.*%u/Exec=\/usr\/share\/%{shortname}\/%{shortname} %U/g' flatpak/%{reverse_dns}.desktop
 sed -i 's/Icon=.*/Icon=%{shortname}/g' flatpak/%{reverse_dns}.desktop
--- a/anda/games/heroic-games-launcher/heroic-games-launcher.spec
+++ b/anda/games/heroic-games-launcher/heroic-games-launcher.spec
@@ -1,105 +1,119 @@
 %global debug_package %{nil}
 %global __provides_exclude ^((libffmpeg[.]so.*)|(lib.*\\.so.*))$
-%global __requires_exclude ^((libffmpeg[.]so.*)|(lib.*\\.so.*))$
+%ifnarch aarch64 
+%global __requires_exclude ^((libffmpeg[.]so.*)|(lib.*\\.so.*)|(.*\\aarch64*\\.so.*))$
+%elifarch aarch64
+%global __requires_exclude ^((libffmpeg[.]so.*)|(lib.*\\.so.*)|(.*\\x86_64*\\.so.*)|(.*\\x86-64*\\.so.*))$
+%endif
 %define _build_id_links none
-%global git_name HeroicGamesLauncher
+%global org_name Heroic-Games-Launcher
+%global git_name %(echo %{org_name} | sed 's/-//g')
+%global reverse_dns com.heroicgameslauncher.hgl
+%global shortname heroic
+%global legendary_version 0.20.36
+%global gogdl_version 1.1.2
+%global nile_version 1.1.2
+%global comet_version 0.2.0
 
-Name:          heroic-games-launcher
+Name:          %{shortname}-games-launcher
 Version:       2.16.1
 Release:       1%?dist
 Summary:       A games launcher for GOG, Amazon, and Epic Games
 License:       GPL-3.0-only AND MIT AND BSD-3-Clause
 URL:           https://heroicgameslauncher.com
-Source0:       https://github.com/Heroic-Games-Launcher/%{git_name}/archive/refs/tags/v%{version}.tar.gz
-Source1:       https://raw.githubusercontent.com/Heroic-Games-Launcher/%{git_name}/refs/heads/main/flatpak/com.heroicgameslauncher.hgl.desktop
-### Makes it actually sign the package, though will say it was skipped first.
-Patch0:        afterPack.diff
-BuildRequires: bsdtar
+BuildRequires: anda-srpm-macros
 BuildRequires: desktop-file-utils
-### Electron builder builds some things with GCC(++) and Make
+### Electron builder builds some things with GCC(++), Git, and Make
 BuildRequires: gcc
 BuildRequires: gcc-c++
-BuildRequires: libxcrypt-compat
+BuildRequires: git
 BuildRequires: make
 BuildRequires: nodejs
 BuildRequires: pnpm
 BuildRequires: python3
+BuildRequires: sed
 Requires:      alsa-lib
-Requires:      atk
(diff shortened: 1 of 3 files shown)


anda/games/heroic-games-launcher/heroic-games-launcher.spec
```diff
@@ -1,105 +1,119 @@
...
 Packager:      Gilver E. <rockgrub@disroot.org>
 
 %description
 Heroic is a Free and Open Source Epic, GOG, and Amazon Prime Games launcher for Linux, Windows, and macOS.
 
 %prep
-%autosetup -n %{git_name}-%{version} -p1
-sed -i 's/Exec=.*%u/Exec=\/usr\/share\/heroic\/heroic %u/g' %{SOURCE1}
-sed -i 's/Icon=.*/Icon=heroic/g' %{SOURCE1}
+rm -rf ./*
```
madonuko: what the heck is this

anda/games/heroic-games-launcher/heroic-games-launcher.spec
```diff
@@ -1,105 +1,119 @@
 %global debug_package %{nil}
 %global __provides_exclude ^((libffmpeg[.]so.*)|(lib.*\\.so.*))$
-%global __requires_exclude ^((libffmpeg[.]so.*)|(lib.*\\.so.*))$
+%ifnarch aarch64 
+%global __requires_exclude ^((libffmpeg[.]so.*)|(lib.*\\.so.*)|(.*\\aarch64*\\.so.*))$
+%elifarch aarch64
+%global __requires_exclude ^((libffmpeg[.]so.*)|(lib.*\\.so.*)|(.*\\x86_64*\\.so.*)|(.*\\x86-64*\\.so.*))$
```
madonuko, line 6: "if not arm, exclude arm, else if arm, exclude x64"
?

anda/games/heroic-games-launcher/update.rhai
```diff
@@ -1 +1,17 @@
-rpm.version(gh("Heroic-Games-Launcher/HeroicGamesLauncher"));
+let v = gh("Heroic-Games-Launcher/HeroicGamesLauncher");
```
madonuko: can you grab the version from https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/blob/v2.15.2/package.json#L3 instead? `gh_rawfile()` doesn't use the api while `gh()` does

anda/games/heroic-games-launcher/heroic-games-launcher.spec
```diff
@@ -1,105 +1,119 @@
...
-%_datadir/applications/heroic.desktop
-%_iconsdir/hicolor/16x16/heroic.png
-%_iconsdir/hicolor/32x32/heroic.png
-%_iconsdir/hicolor/48x48/heroic.png
-%_iconsdir/hicolor/64x64/heroic.png
-%_iconsdir/hicolor/128x128/heroic.png
-%_iconsdir/hicolor/256x256/heroic.png
-%_iconsdir/hicolor/512x512/heroic.png
-%_iconsdir/hicolor/1024x1024/heroic.png
+%license LICENSE
```
madonuko, line 95: 1. I can't find `LICENSE`?
2. what's the difference?

anda/games/heroic-games-launcher/heroic-games-launcher.spec
GildedRoach: If not aarch64 exclude aarch64 libraries from being deps, if aarch64 exclude x86_64 ones. Yeah it's cursed.

What this solves is the reason I noticed  AutoReq on Electron was kinda messed up because RPM got so confused by this thing building aarch64 stuff it expected aarch64 glibc which obviously nothing provides on an x86_64 system (the opposite would also likely be true for aarch64, hence the ifnarch). Can't even disable it building that stuff because it is now just part of the Linux build process.

anda/games/heroic-games-launcher/heroic-games-launcher.spec
GildedRoach: Just the funny remove the buildroot stuff I usually see in Git cloning specs? A lot of the Terra ones do it so I did too. This builds in Mock so it doesn't affect much.

anda/games/heroic-games-launcher/heroic-games-launcher.spec
GildedRoach: It's the license for Legendary which gets fetched and stuck in the build at build time. All this does is take it from the /usr/share folder and move it to the license directory.

anda/games/heroic-games-launcher/update.rhai
GildedRoach: For this since it's the version of the main package, it'd have to be from the main branch but if that always follows the current released version it *should* be doable.

anda/games/heroic-games-launcher/heroic-games-launcher.spec
madonuko: That's fucking cursed

anda/games/heroic-games-launcher/heroic-games-launcher.spec
madonuko: yeah don't do that; old specs do this because `%git_clone` had different behaviour back then

anda/games/heroic-games-launcher/update.rhai
madonuko: oh yeah replace `v2.15.2` with `main`

anda/games/heroic-games-launcher/heroic-games-launcher.spec
GildedRoach: It is absolutely the worst and figuring out a solution to it was horrendous ~~(mostly because I suck at regex)~~.

madonuko (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|f40|[](https://github.com/terrapkg/packages/pull/4002)|
|✅|f41|[](https://github.com/terrapkg/packages/pull/4003)|
|✅|f42|[](https://github.com/terrapkg/packages/pull/4004)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/13927422345) for details

PR #3429, opened 2025-02-19
fix(files): apparmor

madonuko:

PR #4046, opened 2025-03-20
fix: ayatana-indicator-sound

madonuko:


madonuko:
blocked by #3429 (and other stuff, need to rebuild the entire supply chain)

PR #4360, opened 2025-04-20
fix(v4l2loopback): Obsolete separate akmod modules package as this no longer seems necessary

GildedRoach:


madonuko:
can I get an explanation? confusion

PR #4536, opened 2025-04-29
fix: ayatana-indicator-datetime

madonuko:


madonuko:
ok this dependency straight up doesn't exist

GildedRoach:
That's really weird that doesn't exist because `cmake(ECM)` is `extra-cmake-modules` and Fedora does package it.

Maybe just explicit dep it?

Issue #2070, opened 2024-09-12
[Request] gpuopen-amdvlk-drivers


korewaChino:
**What software are you requesting to be packaged?**
https://github.com/GPUOpen-Drivers/AMDVLK

**Describe the motivation**
Official AMD Vulkan drivers as an alternative to Mesa's RADV

**Additional context**
Add any other context about the package request here.

PR #4690, opened 2025-05-05
feat: amdvlk

VirtualFreeEx:
Aims to fulfill package request #2070. Discussion (more of a monologue) on some key issues I have with both the proposal itself and my current implementation of it will be provided below in comments.


VirtualFreeEx:
1. AMDVLK already provides their builds of the driver for RHEL/as an RPM, so is this package _really needed_?
2. DirectXShaderCompiler
    1. (by default) Should we build with external libraries or with libraries from Fedora's repositories?

VirtualFreeEx:
> AMDVLK already provides their builds of the driver for RHEL/as an RPM, so is this package _really needed_?

Specifically on this topic: ~~I believe that the upstream's specfile~~ could be used instead if we don't trust the upstream-provided binaries. Alternatively, could AMDVLK be somehow shipped as the upstream-provided package?

Edit 1: Upstream does not provide a specfile and instead uses CPack to create the package.

madonuko:
we (fyra) are having an ongoing discussion on whether we allow shipping upstream-provided rpms directly, but before we draw a conclusion I think it's not a bad idea to just provide our own rpms anyway

anda/lib/directxshadercompiler/DirectXShaderCompiler.spec
```diff
@@ -0,0 +1,243 @@
...
+# Attribution: https://github.com/gentoo/guru/blob/master/dev-util/DirectXShaderCompiler/DirectXShaderCompiler-1.8.2407.ebuild
+%build
+%cmake \
+    -C ./cmake/caches/PredefinedParams.cmake \
+    -DSPIRV_WERROR=0 \
+    -DLLVM_BUILD_DOCS=0 \
+    -DLLVM_BUILD_TOOLS=0 \
+    -DSPIRV_BUILD_TESTS=0 \
+    -DLLVM_ENABLE_WERROR=0 \
+    -DBUILD_SHARED_LIBS=OFF \
```
madonuko, line 76: interesting, I wonder why is this off but we still have the shared libs?

anda/lib/directxshadercompiler/DirectXShaderCompiler.spec
```diff
@@ -0,0 +1,243 @@
...
+# Files.
+%files tools
+%{_libdir}/%{name}/bin/
+%{_bindir}/dxa
+%{_bindir}/dxc
+%{_bindir}/dxl
+%{_bindir}/dxopt
+%{_bindir}/dxr
+%{_bindir}/dxv
+
```
madonuko: why putting everything into the subpackage when you can just put them into the main package?

anda/lib/directxshadercompiler/DirectXShaderCompiler.spec
```diff
@@ -0,0 +1,243 @@
...
+%files libs
+%{_libdir}/%{name}/lib/libdxcompiler.so
+%{_libdir}/%{name}/lib/libdxil.so
+
+%files static
+%{_libdir}/%{name}/lib/libHLSLTestLib.a
```
madonuko: could just do `*.a`?

anda/lib/directxshadercompiler/DirectXShaderCompiler.spec
```diff
@@ -0,0 +1,243 @@
...
+    -C ./cmake/caches/PredefinedParams.cmake \
+    -DSPIRV_WERROR=0 \
+    -DLLVM_BUILD_DOCS=0 \
+    -DLLVM_BUILD_TOOLS=0 \
+    -DSPIRV_BUILD_TESTS=0 \
+    -DLLVM_ENABLE_WERROR=0 \
+    -DBUILD_SHARED_LIBS=OFF \
+    -DLLVM_VERSION_SUFFIX=dxc \
+    -DSPIRV_WARN_EVERYTHING=0 \
+    -DCMAKE_INSTALL_PREFIX="%{_libdir}/%{name}" \
```
madonuko: why not just `%_prefix`?

GildedRoach: I looked at the files. This should be built into the prefix. This does also include CMake files which are currently in the wrong location meaning CMake probably cannot find them and which also belong in the devel package.

madonuko: this

anda/lib/directxshadercompiler/DirectXShaderCompiler.spec
VirtualFreeEx: i.. dont really have an idea to be honest.

anda/lib/directxshadercompiler/DirectXShaderCompiler.spec
VirtualFreeEx: fair point, subpackage removed in local tree, will be pushed later

anda/lib/directxshadercompiler/DirectXShaderCompiler.spec
VirtualFreeEx: done

anda/lib/directxshadercompiler/DirectXShaderCompiler.spec
VirtualFreeEx: because it would possibly conflict with files from llvm's files

GildedRoach: The package is not usable by CMake *or* LLVM this way. This has CMake files and compiler files that must be in the intended locations or they can't be used.

VirtualFreeEx: removed the CMake files (i dont think i was correct in the first place with including them)

i dont think that llvm would use this package by itself since this package happens to be a fork of llvm and specifically clang afaik, not something that uses it at runtime

korewaChino:
AMDVLK is dead, long live RADV

https://www.phoronix.com/news/AMDVLK-Discontinued

GildedRoach:
> AMDVLK is dead, long live RADV
> 
> https://www.phoronix.com/news/AMDVLK-Discontinued

While that is true, some video and other editing software do not accept or work with Mesa. It's a frustrating situation.

PR #4857, opened 2025-05-16
add: Terra WINE

GildedRoach:
I know there will understandably be questions about what in the world is happening with these specs so let me answer a few.

- About 80% of the specs are taken directly from WineHQ. That INCLUDES the global definitions. I didn't change them.
- MinGW GCC does NOT support most hardening flags. This is an issue upstream Fedora has as well, their WINE specs are also a mess.
- Fedora's Red Hat hardening is known to break WINE, Fedora also notes this.
- Why does WineHQ install to a different directory than root and symlink everything? Absolutely no idea. This did, however, allow me to just rename the `winehq-wine` packages to `terra-` since there were changes in these.
- Why were these added? I originally notified Lea, but the reason is Fedora's WINE stack does not update fast enough to keep up with our Mesa updates. 
  - Pulling from WineHQ also let me do different branches of WINE. Fedora *only* provides Staging. While this is generally the recommended WINE to use, for many reasons people would also use dev or stable instead.

I will also need to update the Terra Mesa repo required for builds on the other branches but the alternative to that was like four different PRs.


GildedRoach:
`wine-dxvk` build failures are expected, these can be moved to a different PR if needed. But yay tarball integrity checked now.

korewaChino (approved):

madonuko (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|f41|[](https://github.com/terrapkg/packages/pull/4894)|
|✅|el10|[](https://github.com/terrapkg/packages/pull/4896)|
|✅|f42|[](https://github.com/terrapkg/packages/pull/4897)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/15111981512) for details

PR #4916, opened 2025-05-19
fix(ffmpeg): Move to Extras due to library deps

GildedRoach:
Was looking at build logs again (don't ask) and noticed this library deps on `fdk-aac` which we only have in Extras.

This *shouldn't* break any upgrade paths at all since the package has a different name from Fedora's, and should be able to be swapped to in the same way Fusion's is.

### DIFF
--- a/anda/multimedia/ffmpeg/anda.hcl
+++ b/anda/multimedia/ffmpeg/anda.hcl
@@ -7,5 +7,6 @@ project pkg {
     labels {
         updbranch = 1
         mock = 1
+        subrepo = "extras"
     }
 }


madonuko:
does this work without extras? the use of extras packages during build time doesn't imply it requires runtime dependencies from extras

madonuko:
ah nvm misunderstood

madonuko (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|el10|[](https://github.com/terrapkg/packages/pull/4921)|
|✅|f42|[](https://github.com/terrapkg/packages/pull/4922)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/15129959936) for details

PR #4951, opened 2025-05-21
chore(kmod-nvidia): Update for Alma production branch

GildedRoach:
Can be merged once we switch.

If a better way to track RHEL kernel versions becomes apparent I will adapt that instead.


GildedRoach:
In theory with this change I can just redefine the global every time the kernel version bumps but I'll wait for Mado's wacky release bump solution because that'd be useful here.

anda/system/nvidia/kmod-nvidia/update.rhai
```diff
@@ -3,8 +3,12 @@ import "andax/nvidia.rhai" as nvidia;
...
-let kr = find("specrelease ([\\d.]+)", rawfile, 1);
-open_file("anda/system/nvidia/kmod-nvidia/KERNEL_VERSION.txt", "w").write(kv);
-open_file("anda/system/nvidia/kmod-nvidia/KERNEL_RELEASE.txt", "w").write(kr);
+let releasever = [];
+for matches in find_all(`([\d]+\.[\d]+)/`, get("https://repo.almalinux.org/almalinux/")) {
+        releasever += matches[1].parse_float();
+}
+releasever.dedup();
+releasever.sort();
+let kver = find_all(`kernel-([\d.]+.*?)\.el.*?\.x86_64\.rpm`, get(`https://repo.almalinux.org/almalinux/${releasever[releasever.len()-1]}/BaseOS/x86_64/os/Packages/`));
```
madonuko: what is `.*?`, isn't that the same as `.*`?

GildedRoach: Not quite, `?` converts it to "non-greedy" regex.

Otherwise it matches the entire HTML line instead of stopping after the next character given is found.

madonuko: "non-greedy" regexes don't exist in rust

madonuko: nvm it does, I misremembered or something then

GildedRoach: You can check it locally, it did fix the matching.

If that's not the reason why I'm not sure why.

GildedRoach: > nvm it does, I misremembered or something then

akskskskskdkdkdk

GildedRoach:
WE ARE SO BACK

PR #5005, opened 2025-05-24
deprecate vvenc and use rpmfusion upstream insted

solomoncyj:
https://admin.rpmfusion.org/pkgdb/package/free/vvenc/


madonuko:
I thought vvenc was broken in rpmfusion?

solomoncyj:
wdym?

madonuko:
we added vvenc because it was broken on rpmfusion

GildedRoach:
Why would we do this...? Furthermore, we *can't* do this because FFmpeg requires it to build and we do not use RPM Fusion repos in our builds.

madonuko:
honestly sorry I don't remember, @RockGrub

GildedRoach:
> honestly sorry I don't remember, @RockGrub

It was added due to being a build dep for FFmpeg.

solomoncyj:
:facepalm:

GildedRoach:
RPM Fusion isn't our upstream, Fedora is, so I don't understand why this is a suggestion.

We also want to ideally work on systems without RPM Fusion too, which we can't do if we do this, even the inability to build FFmpeg itself (which we *also* need for some of our builds that need FFmpeg to build) aside.

Personally I think this is a bad idea.

solomoncyj:
this will create conflicting package problem if a use has both repos enabled

GildedRoach:
> this will create conflicting package problem if a use has both repos enabled

I have both and it hasn't. But if that is a concern, the solution is to move it to Extras, not remove it.

madonuko:
`dnf in vvenc` is giving me the package from rpmfusion-free-updates

edit: nvm this package isn't in terra41

GildedRoach:
> `dnf in vvenc` is giving me the package from rpmfusion-free-updates
> 
> edit: nvm this package isn't in terra41

I tested on a 42 box (a few of them) and had no issues myself, but yeah, we can probably move this to Extras if there's concern since we also moved FFmpeg.

madonuko:
honestly we should reconsider adding a repo priority to terra itself and remove, but obviously that's a lea question

for now I think moving this to extras would be fine (remember to satm it tho)

GildedRoach:
> honestly we should reconsider adding a repo priority to terra itself and remove, but obviously that's a lea question
> 
> for now I think moving this to extras would be fine (remember to satm it tho)

If you mean lowering Terra priority, Jade did bring up a good point that Fedora likes that we don't carry overlapping packages with them in our base repo (priority aside), and lowering repo priority does worry me just because that is a *lot* of packages we probably cannot individually test if that will cause issues on and there's been some bizarre DNF behavior as of late.

We do need this for FFmpeg either way. As long as it exists in one of our builder repos that should be fine.

nothingneko:
number two rule is don't fuck with fedora—anything that overrides fedora is either in extras or not allowed

number one rule is don't break systems, don't break systems when you do this rpmfusion stuff it keeps happening please be careful

nothingneko:
number two rule is don't fuck with fedora—anything that overrides fedora is either in extras or not allowed

number one rule is don't break systems, don't break systems when you do this rpmfusion stuff it keeps happening please be careful

GildedRoach:
> number two rule is don't fuck with fedora—anything that overrides fedora is either in extras or not allowed
> 
> number one rule is don't break systems, don't break systems when you do this rpmfusion stuff it keeps happening please be careful

This PR will break systems far more than us just having VVenc because there are users on our FFmpeg package, but like I said it *should* be fine to move to extras.

PR #5404, opened 2025-06-14
feat(wine-dxvk): Dep on any WINE

GildedRoach:


anda/system/wine-dxvk/wine-dxvk.spec
```diff
@@ -143,8 +142,7 @@ Requires:       terra-wine-dxvk-d3d8(x86-32) = %{version}-%{release}
 %{summary}
 
 %prep
-%autosetup -n dxvk-%{version} -p1 -a1
-rm -rf subprojects/libdisplay-info && mv libdisplay-info-%{libdisplay_commit} subprojects/libdisplay-info
+%git_clone %{url}.git %{version}
```
madonuko:
```suggestion
%git_clone %url v%version
```
magic

GildedRoach: IDK why it's showing this as still being broken to you because I fixed it hours ago. :sob:

madonuko: idrk…? aa

korewaChino:
funny idea: what if we just require `/usr/bin/wine`? Any package that provides that binary should work and it would scale

GildedRoach:
> funny idea: what if we just require `/usr/bin/wine`? Any package that provides that binary should work and it would scale

That's not the issue, the issue is the alternatives post scripts.

PR #5652, opened 2025-06-27
fix(ffmpeg): openh264 is not provided in 32-bit

GildedRoach:


GildedRoach:
I forgot about the DNF bug. :)

madonuko:
real
also edit anda.hcl for arches list with 32 bit

madonuko:
oh nvm it's already there

GildedRoach:
Actually LCDEfdufsdfsdjkfsdjk has changes too I should deal with that first probably

GildedRoach:
> Actually LCDEfdufsdfsdjkfsdjk has changes too I should deal with that first probably

Nevermind the changes Negativo did are kinda weird and not worth replicating we good.

GildedRoach:
I hate noopenh264 it is such a stupid package.

GildedRoach:
Actually it appears `noopenh264` 32-bit can be safely installed alongside 64-bit actual openh264 so maybe this pull request isn't needed?

PR #5659, opened 2025-06-28
add: lan-mouse

ayykamp:
Package [lan-mouse](https://github.com/feschber/lan-mouse) for Terra.

Partially based on https://build.opensuse.org/projects/home:dr-yak/packages/lan-mouse/files/lan-mouse.spec?expand=1

PR #5679, opened 2025-06-28
fix(rnote): remove unnecessary line 

ayykamp:
See https://github.com/terrapkg/packages/pull/5659#discussion_r2173233596


ayykamp:
Oh you were faster @RockGrub :)

PR #5762, opened 2025-07-04
Add: flashrom-cros-devel

Owen-sz:


Owen-sz:
Needed dependency for another chromiumOS package

PR #5778, opened 2025-07-05
fix(ci): grep version properly

madonuko:


madonuko:
this should hopefully fix the graalvm issue…

nothingneko (approved):

PR #5806, opened 2025-07-12
fix(curl-impersonate-chrome): rm patches

madonuko:


GildedRoach:
I got this to build in the CI before if I used Mock WTF

PR #5810, opened 2025-07-12
fix: python-pip-system-certs

madonuko:


madonuko:
seems to be upstream

GildedRoach:
The `%pyproject` macros are (or at least should be) correct on all branches for this package. It uses a `pyproject.toml`.

madonuko:
yeah I was like maybe the old macros would do something else

nope, license is still messed up according to both macros

PR #6024, opened 2025-08-08
add: bitwarden-desktop

madonuko:


pointydev:
For reference, openSUSE uses this spec file for their desktop build: https://build.opensuse.org/projects/devel:languages:nodejs/packages/bitwarden/files/bitwarden.spec

GildedRoach:
Is this Electron? If so it needs to be in Mock or it will intermittently or completely fail especially if it's using NPM.

It's some stupid stuff with NPM's runtime.

madonuko:
it's not the usual error we see from things like vesktop tho, the dir it's looking for just somehow doesn't exist

PR #6142, opened 2025-08-23
Update nekoray

qr243vbi:
Original nekoray project is officially abandoned and archived, and unofficial one also dissapeared, so i decided to create nekoray fork as continuation of that legendary project


madonuko:
are there any major adoptions?
also if your name is not approved by the previous upstream you are potentially violating their trademark rights

BOplaid:
Instead of making your own fork, go contribute to [Throne](https://github.com/throneproj/Throne)

PR #6403, opened 2025-09-16
fix: ghostty

madonuko:


GildedRoach:
This is incorrect. The nightly changes should be imported instead.

Also Zig is too old now on F41...so...

PR #6434, opened 2025-09-20
chore: rebuild ghostty 1.1.3

madonuko:
See #6433 
This is an attempt to bring back ghostty 1.1.3


madonuko:
right, zig is breaking

PR #6671, opened 2025-10-03
fix: scrcpy

Owen-sz:


madonuko (commented):
bump release

korewaChino:
I wonder what if we just build it as a separate package and disable BRP on the binary so x86 systems can install it, so we get this cursed dependency on `scrcpy-server.aarch64`

PR #7044, opened 2025-11-02
Build Crystal from source, split out shards pkg

carlhoerberg:
Build Crystal and shards from source, building dynamically linked binaries.

Separate the crystal and shards packages, they have different versions and update schedules.

Remove architecture restrictions.

Introduce a bootstrap problem, as Crystal is built using Crystal.

### DIFF
--- a/anda/langs/crystal/crystal/crystal.spec
+++ b/anda/langs/crystal/crystal/crystal.spec
@@ -1,24 +1,24 @@
 %bcond bootstrap 1
 %global bootstrap_version 1.17.1
 
-Name:			  crystal
-Version:		1.18.2
-Release:		2%?dist
-Summary:    A general-purpose, object-oriented programming language
-License:		Apache-2.0
-URL:			  https://crystal-lang.org/
-Source0:    https://github.com/crystal-lang/crystal/archive/%version.tar.gz
+Name:          crystal
+Version:       1.18.2
+Release:       2%?dist
+Summary:       A general-purpose, object-oriented programming language
+License:       Apache-2.0
+URL:           https://crystal-lang.org/
+Source0:       https://github.com/crystal-lang/crystal/archive/%version.tar.gz
 %if %{with bootstrap}
-Source1:    https://dev.alpinelinux.org/archive/crystal/crystal-%{bootstrap_version}-%{_arch}-alpine-linux-musl.tar.gz
+Source1:       https://dev.alpinelinux.org/archive/crystal/crystal-%{bootstrap_version}-%{_arch}-alpine-linux-musl.tar.gz
 %else
-BuildRequires:  crystal
+BuildRequires: crystal
 %endif
-BuildRequires:  gcc gcc-c++ make gc-devel llvm-devel
-BuildRequires:  pcre2-devel libyaml-devel libffi-devel
-Requires:       gcc pkgconfig gc-devel
-Requires:       pcre2-devel openssl-devel zlib-devel
-Requires:       libyaml-devel libxml2-devel gmp-devel
-Suggests:       shards
+BuildRequires: gcc gcc-c++ make gc-devel llvm-devel
+BuildRequires: pcre2-devel libyaml-devel libffi-devel
+Requires:      gcc pkgconfig gc-devel
+Requires:      pcre2-devel openssl-devel zlib-devel
+Requires:      libyaml-devel libxml2-devel gmp-devel
+Suggests:      shards
 
 %description
 Crystal is a programming language with the following goals:
--- a/anda/langs/crystal/crystal/crystal.spec
+++ b/anda/langs/crystal/crystal/crystal.spec
@@ -29,7 +29,7 @@ Crystal is a programming language with the following goals:
 - Compile to efficient native code
 
 %prep
-%autosetup -v
+%autosetup
 %if %{with bootstrap}
 %setup -b 1
 %endif
--- a/anda/langs/crystal/crystal/crystal.spec
+++ b/anda/langs/crystal/crystal/crystal.spec
@@ -1,4 +1,5 @@
-%define debug_package %nil
+%bcond bootstrap 1
+%global bootstrap_version 1.17.1
 
 Name:			  crystal
 Version:		1.18.2
@@ -7,7 +8,12 @@ Summary:    A general-purpose, object-oriented programming language
 License:		Apache-2.0
 URL:			  https://crystal-lang.org/
 Source0:    https://github.com/crystal-lang/crystal/archive/%version.tar.gz
-BuildRequires:  crystal gcc gcc-c++ make gc-devel llvm-devel
+%if %{with bootstrap}
+Source1:    https://dev.alpinelinux.org/archive/crystal/crystal-%{bootstrap_version}-%{_arch}-alpine-linux-musl.tar.gz
+%else
+BuildRequires:  crystal
+%endif
+BuildRequires:  gcc gcc-c++ make gc-devel llvm-devel
 BuildRequires:  pcre2-devel libyaml-devel libffi-devel
 Requires:       gcc pkgconfig gc-devel
 Requires:       pcre2-devel openssl-devel zlib-devel
@@ -23,13 +29,21 @@ Crystal is a programming language with the following goals:
 - Compile to efficient native code
 
 %prep
+%if %{with bootstrap}
+%setup -q -b 1
+%else
 %setup -q
+%endif
 
 %build
-make release=1 interpreter=1 LDFLAGS="%{build_ldflags}" CRYSTAL_CONFIG_LIBRARY_PATH=%{_libdir}/crystal
+%if %{with bootstrap}
+# Use bootstrap crystal binary
+export PATH="%{_builddir}/crystal-%{bootstrap_version}/bin:$PATH"
+%endif
+%make_build release=1 interpreter=1 LDFLAGS="%{build_ldflags}" CRYSTAL_CONFIG_LIBRARY_PATH=%{_libdir}/crystal
 
 %install
-make install DESTDIR=%{buildroot} PREFIX=%{_prefix}
+%make_install PREFIX=%{_prefix}
 
 %files
 %license %{_datadir}/licenses/crystal/LICENSE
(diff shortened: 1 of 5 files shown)


GildedRoach:
For the last point, this is basically why RPM has `%bcond bootstrap`. Maybe worth looking into?

carlhoerberg:
Ok, yes, so create a new `crystal-boostrap` package which uses the current way of downloading precompiled binaries from github? and then in this package

```spec
%bcond bootstrap 0
%if %{with bootstrap}
BuildRequires: crystal-bootstrap
%else
BuildRequires: crystal
%endif
```

or if the `--with bootstrap` flag is passed the build step downlaods and uses the prebuild binaries?  

```spec
%bcond bootstrap 0
%build
%if %{with bootstrap}
...download precompiled binaries...
%endif
```

GildedRoach:
> Ok, yes, so create a new `crystal-boostrap` package which uses the current way of downloading precompiled binaries from github? and then in this package
> 
> ```python
> %bcond bootstrap 0
> %if %{with bootstrap}
> BuildRequires: crystal-bootstrap
> %else
> BuildRequires: crystal
> %endif
> ```
> 
> or more something like
> 
> ```python
> %bcond bootstrap 0
> %build
> %if %{with bootstrap}
> ...download precompiled binaries...
> %endif
> ```

That works, technically can also do it in the same spec but a separate one may be easier if it needs to regularly be bootstrapped with the latest or second to latest version.

anda/langs/crystal/crystal/crystal.spec
```diff
@@ -1,57 +1,49 @@
...
-mkdir -p %buildroot/usr/bin
-mkdir -p %buildroot/usr/share
-mkdir -p %buildroot/usr/lib/crystal
-install -Dm755 bin/* %buildroot/usr/bin/
-cp -r share/* %buildroot/usr/share/
-cp -r lib/crystal/* %buildroot/usr/lib/crystal/
-
-%fdupes %buildroot%_datadir/crystal/src/lib_c/
-
+make install DESTDIR=%{buildroot} PREFIX=%{_prefix}
```
Owen-sz:
```suggestion
%make_install DESTDIR=%{buildroot} PREFIX=%{_prefix}
```

anda/langs/crystal/crystal/crystal.spec
```diff
@@ -1,57 +1,49 @@
...
+- Be able to call C code by writing bindings to it in Crystal
+- Have compile-time evaluation and generation of code, to avoid boilerplate code
+- Compile to efficient native code
 
 %prep
-%autosetup -n crystal-%version-1
+%setup -q
 
 %build
+make release=1 interpreter=1 LDFLAGS="%{build_ldflags}" CRYSTAL_CONFIG_LIBRARY_PATH=%{_libdir}/crystal
```
Owen-sz: %make_build?

anda/langs/crystal/shards/shards.spec
```diff
@@ -0,0 +1,34 @@
...
+
+
+%prep
+%setup -q
+
+%build
+make release=1 FLAGS="--link-flags=\"%{build_ldflags}\""
+
+%install
+make install DESTDIR=%{buildroot} PREFIX=%{_prefix}
```
Owen-sz: Macros on here too?

Owen-sz (changes requested):

anda/langs/crystal/crystal/crystal.spec
Owen-sz: Also, why remove %autosetup?

anda/langs/crystal/crystal/crystal.spec
```diff
@@ -1,57 +1,49 @@
...
+Release:		2%?dist
+Summary:    A general-purpose, object-oriented programming language
 License:		Apache-2.0
-URL:			https://crystal-lang.org/
-Source0:		https://github.com/crystal-lang/crystal/releases/download/%version/crystal-%version-1-linux-x86_64-bundled.tar.gz
-ExclusiveArch:	x86_64
-BuildRequires:	rpm_macro(fdupes)
+URL:			  https://crystal-lang.org/
+Source0:    https://github.com/crystal-lang/crystal/archive/%version.tar.gz
+BuildRequires:  crystal gcc gcc-c++ make gc-devel llvm-devel
```
madonuko: unspoken rule in terra, but we prefer not having recursive build dependencies, because that makes it impossible to bootstrap this package automatically during branch-offs.

This means you must either always build crystal from the ground up (or download prebuilt binaries somewhere then build it, but of course the arch selection is limited), or use some sort of `pre.rhai` setup to determine if a `crystal` package is available.

~~The latter can be done like this:~~
1. create `pre.rhai` in this directory, with:
```ts
let stdout = sh("dnf rq crystal", #{"stdout": "piped"}).ctx.stdout;
stdout.trim();
if stdout == "" {
  write("1", "crystal.txt");
}
```
2. in the spec file, use [`%{exists:crystal.txt}`](https://developer.fyralabs.com/rpm/macros#exists) to check for the file existence, e.g.:
```
%if %{exists:crystal.txt}
# no crystal in dnf, bootstrap required
%else
BuildRequires: crystal
%endif
```

(Alternatively I don't know if that works but you can try using `%(dnf rq crystal)` directly inside the spec file.)

Another way which has already been mentioned by roachy is to create a bootstrap package, and you can see `x264-bootstrap` and `x264` as an example.

For the "build from the ground up" way, see `nim` and `nim-nightly`.

anda/langs/crystal/crystal/crystal.spec
```diff
@@ -1,57 +1,49 @@
...
-/usr/share/man/man5/shard.yml.5.gz
-
-%files devel
-/usr/lib/crystal/
+%license %{_datadir}/licenses/crystal/LICENSE
+%{_bindir}/crystal
+%{_datadir}/crystal
+%{_datadir}/zsh/site-functions/_crystal
+%{_datadir}/bash-completion/completions/crystal
+%{_datadir}/fish/vendor_completions.d/crystal.fish
```
madonuko, line 54: We haven't updated this spec file in a while, but we have new guidelines and macros now.
We can remove these 3 lines and put `%pkg_completion -Bfz` before `%prep`

anda/langs/crystal/crystal/crystal.spec
```diff
@@ -1,57 +1,49 @@
 %define debug_package %nil
```
madonuko: if built from source probably doesn't need this

madonuko (commented):
thank you for the pr

anda/langs/crystal/crystal/crystal.spec
carlhoerberg: Or would a crystal-bootstrap package with prebuilt binaries be preferable?

madonuko: Then you can only support x86_64. I don't think they provide binaries for other architectures.

anda/langs/crystal/crystal/crystal.spec
carlhoerberg: With it I get: 
```
rpmbuild │ error: Installed (but unpackaged) file(s) found:
rpmbuild │    /usr/share/bash-completion/completions/crystal
rpmbuild │    /usr/share/fish/vendor_completions.d/crystal.fish
rpmbuild │    /usr/share/zsh/site-functions/_crystal
```

anda/langs/crystal/crystal/crystal.spec
carlhoerberg: Updated with bootstrap static binaries from alpine... not sure what your stand is on that, the benefit is that they provide aarch64 static binaries that we can use for bootstrapping, in addition to the x86_64..

anda/langs/crystal/crystal/crystal.spec
madonuko: did you forget to add `%pkg_completion -Bfz`? that should add the files automatically. You might need to put it before / after %description, not sure.

madonuko:
I think you need to add the directory you've downloaded that contains the crystal binary into `PATH`

carlhoerberg:
> I think you need to add the directory you've downloaded that contains the crystal binary into `PATH`

fixed

anda/langs/crystal/crystal/crystal.spec
carlhoerberg: not possible before `%description`. between `%description` and `%prep` yielded this error

anda/langs/crystal/shards/shards.spec
```diff
@@ -0,0 +1,31 @@
...
+Version:       0.19.1
+Release:       1%?dist
+Summary:       Dependency manager for the Crystal language 
+License:       Apache-2.0
+URL:           https://crystal-lang.org/
+Source0:       https://github.com/crystal-lang/shards/archive/refs/tags/v%version.tar.gz
+BuildRequires: crystal make gcc
+Suggests:      git make
+Supplements:   crystal
+
```
Owen-sz, line 13: Could you also add yourself as the `Packager:`?

https://developer.fyralabs.com/terra/policy#miscellaneous-packaging-guidelines

Thanks again for the PR!

Owen-sz (changes requested):

carlhoerberg:
Not sure why building `shards` failed, have `crystal` as builddeps, and `crysta` got `Depends: gcc`, still building `shards` failed with `gcc not found`

anda/langs/crystal/shards/shards.spec
carlhoerberg: Fixed! Thanks

Owen-sz:
> Not sure why building `shards` failed, have `crystal` as builddeps, and `crysta` got `Depends: gcc`, still building `shards` failed with `gcc not found`

Got it to build locally. Try adding these to the `BuildRequires`: gcc gc-devel libyaml-devel pcre-devel
should be good then!

Owen-sz:
aarch64 shards fails due to aarch64 crystal not existing yet. x86_64 builds fine now so aarch64 should be good too.

Owen-sz (approved):

madonuko:
thank you very much!

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|f41|[](https://github.com/terrapkg/packages/pull/7133)|
|✅|el10|[](https://github.com/terrapkg/packages/pull/7134)|
|✅|f42|[](https://github.com/terrapkg/packages/pull/7135)|
|✅|f43|[](https://github.com/terrapkg/packages/pull/7136)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/19093064349) for details

PR #7316, opened 2025-11-12
chore: Move packages to multimedia repo

GildedRoach:


GildedRoach:
Ironically this repo itself will solve the current build failure.

anda/multimedia/gstreamer1/gstreamer1-plugins-ugly/anda.hcl
```diff
@@ -2,9 +2,10 @@ project pkg {
   arches = ["x86_64", "aarch64", "i386"]
   rpm {
     spec = "gstreamer1-plugins-ugly.spec"
+    extra_repos = ["https://repos.fyralabs.com/terrarawhide-multimedia"]
```
korewaChino: wonder if we can substitute this with a template so it works on all versions

lleyton (approved):

PR #7398, opened 2025-11-15
fix (magic-wormhole): add version remove patch

Owen-sz:


pointydev:
Turns out the issue pointed out in https://github.com/magic-wormhole/magic-wormhole/issues/680 is *not* Windows specific, after testing the artifact for this pull I'm presented with a similar error to the one in the issue after running `wormhole send [file]`.

PR #7600, opened 2025-11-23
fixup: Adjust fdk-aac spec on top of #7597

DekoDX:
I just checked the artifacts of the last pr and the i686 build was building an x86_64 package instead. I need to be certain this builds right before this gets merged.

Deeply apologize. This won't happen again


DekoDX:
Not certain why this wouldn't be building. I think I have the wrong approach. I'll figure this out later :/

PR #7872, opened 2025-12-02
backport: chore: update stardust telescope to reflect upstream changes

Owen-sz:
Updated commit hash and date, and changed release number.

### DIFF
--- a/anda/stardust/telescope/stardust-telescope.spec
+++ b/anda/stardust/telescope/stardust-telescope.spec
@@ -1,12 +1,10 @@
-%define debug_package %nil
-
-%global commit e33764c69179e35b60ad03931544a87357e1e81f
-%global commit_date 20250413
+%global commit 2c514175efc53be985d7d092601f158b38dc29b4
+%global commit_date 20251201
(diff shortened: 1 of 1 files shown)


anda/stardust/telescope/stardust-telescope.spec
```diff
@@ -1,12 +1,10 @@
...
-%global commit e33764c69179e35b60ad03931544a87357e1e81f
-%global commit_date 20250413
+%global commit 2c514175efc53be985d7d092601f158b38dc29b4
+%global commit_date 20251201
 %global shortcommit %(c=%{commit}; echo ${c:0:7})
 
 Name:           stardust-xr-telescope
 Version:        %commit_date.git~%shortcommit
-Release:        2%?dist
+Release:        1%?dist
```
metcya, line 7: because the version is changing do we need to bump the release here?

Owen-sz: No, version trumps release, and since the commit date is higher it will replace this package just fine.
Whenever a version is bumped the release should be reset to 1

metcya (approved):

PR #7903, opened 2025-12-03
chore: update stardust-server to reflect upstream changes (PART 1)

Owen-sz:


Owen-sz:
Ok this *should* be fine now

PR #7913, opened 2025-12-03
chore(rpcs3): Attempt to fix build on F41

GildedRoach:
Hack for until we drop F41.


GildedRoach:
Unfortunate, but it's a project set build flag.

PR #854, opened 2023-12-17
Add Jellyfin MPV Shim

korewaChino:

PR #8152, opened 2025-12-07
add: jellyfin-mpv-shim

### LINKED ISSUE #853: [Request] python3-jellyfin-mpv-shim
https://github.com/jellyfin/jellyfin-mpv-shim

**Describe the motivation**
While this package already has a Flatpak build on flathub, some people might still prefer using their already existing local MPV config with Jellyfin. The shim uses a dedicated MPV build with a separate configuration.

Configuration is also easier to locate when not using the Flatpak version.

**Additional context**
Is a Python package, depends on tkinter, python3-webview, and MPV itself as runtime dependency

metcya:
closes #853 and supersedes #854


metcya:
depends on the following being packaged
- [x] jellyfin-apiclient-python
- [x] python-mpv-jsonipc

metcya:
flatpak works fine for this actually,,,

PR #8343, opened 2025-12-14
fix: use go macros for soft-serve

arbormoss:
This makes soft-serve use go macros for building.

PR #8348, opened 2025-12-14
fix: use go macros for soft-serve

arbormoss:
This refactors soft-serve to use go rpm macros.


Owen-sz:
Isn't this what https://github.com/terrapkg/packages/pull/8343 did?

arbormoss:
oh yep lol

PR #8442, opened 2025-12-18
add: pixi

### LINKED ISSUE #8409: [Request] pixi
`pixi`
https://pixi.sh/
https://github.com/prefix-dev/pixi

**Describe the motivation**
I would like to be able to install this package from DNF rather than an external package manager like Homebrew or Nixpkgs.

**Additional context**
Although it is available on cargo.io, that version is way too behind. It's recommended to clone to source and build it using `cargo build` or install with `cargo install`.

Also, I tried compiling this from source, and it always kept failing at the last step. I hope whoever tries to compile pixi next has better luck than I do.

metcya:
resolves #8409

### DIFF
--- a/anda/system/pixi/anda.hcl
+++ b/anda/system/pixi/anda.hcl
@@ -1,0 +1,5 @@
+project pkg {
+  rpm {
+    spec = "pixi.spec"
+  }
+}
--- a/anda/system/pixi/anda.hcl
+++ b/anda/system/pixi/anda.hcl
@@ -2,7 +2,4 @@ project pkg {
   rpm {
     spec = "pixi.spec"
   }
-  label {
-    large = 1
-  }
 }
--- a/anda/system/pixi/anda.hcl
+++ b/anda/system/pixi/anda.hcl
@@ -2,4 +2,7 @@ project pkg {
   rpm {
     spec = "pixi.spec"
   }
+  label {
+    large = 1
+  }
 }
--- a/anda/system/pixi/pixi.spec
+++ b/anda/system/pixi/pixi.spec
@@ -1,0 +1,44 @@
+Name:           pixi
+Version:        0.62.0
+Release:        1%{?dist}
+Summary:        A cross-platform, multi-language package manager
+License:        BSD-3-Clause
+URL:            https://pixi.sh
+Source:         https://github.com/prefix-dev/pixi/archive/refs/tags/v%{version}.tar.gz
+Packager:       metcya <metcya@gmail.com>
+
+BuildRequires:  anda-srpm-macros
+BuildRequires:  cargo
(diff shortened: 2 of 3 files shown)


Owen-sz:
Why `%dnl %cargo_license_online`?

Owen-sz:
You need to first do %cargo_license_summary_online

So do this:
```
%cargo_license_summary_online
%{cargo_license_online} > LICENSE.dependencies
```

GildedRoach:
> You need to first do %cargo_license_summary_online
> 
> So do this:
> 
> ```
> %cargo_license_summary_online
> %{cargo_license_online} > LICENSE.dependencies
> ```

That actually only outputs a summary of the licenses, it's only for the packager to know what to put in the license field and is not actually necessary for the build.

Owen-sz:
Oh thought it prepped it... why was this failing then?
Might have a look locally

GildedRoach:
> Oh thought it prepped it... why was this failing then? Might have a look locally

It looks like it gets stuck which if it actually is, that *might* be the same bug Zed hits with `cargo tree`. I see some of the same Crates, might be relevant?

Owen-sz:
> > Oh thought it prepped it... why was this failing then? Might have a look locally
> 
> 
> 
> It looks like it gets stuck which if it actually is, that *might* be the same bug Zed hits with `cargo tree`. I see some of the same Crates, might be relevant?

Odd... I guess it's not strictly needed.

GildedRoach:
> > > Oh thought it prepped it... why was this failing then? Might have a look locally
> > 
> > 
> > It looks like it gets stuck which if it actually is, that _might_ be the same bug Zed hits with `cargo tree`. I see some of the same Crates, might be relevant?
> 
> Odd... I guess it's not strictly needed.

The bug in Zed was the `--no-dedupe` flag so could do something like my workaround in Zed if that was the issue with it. https://github.com/korewaChino/packages/blob/71f090b172a34b3b62c994b6cfac8d951a11d4ea/anda/devs/zed/stable/zed.spec#L89

Owen-sz (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|f41|[](https://github.com/terrapkg/packages/pull/8463)|
|✅|el10|[](https://github.com/terrapkg/packages/pull/8464)|
|✅|f42|[](https://github.com/terrapkg/packages/pull/8465)|
|✅|f43|[](https://github.com/terrapkg/packages/pull/8466)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/20351307528) for details

PR #8537, opened 2025-12-22
add: uxn

### LINKED ISSUE #8538: Uxn Versioning
### Background

This is for discussion of how to version [uxn](https://100r.co/site/uxn.html).

See this [Discord message](https://discord.com/channels/957104883490775061/1209327267852058704/1452519075187261543) for the full versioning issue or this [backup of the conversion](https://paste.sr.ht/~arbormoss/518575d33e33414e803df1433abcbeaf2576c823) for the initial conversation on the versioning issue.

## Uxn Versioning

I am currently thinking just have a nightly release for the continuously updated version.

I am not sure what to do about the frozen/stable version. I do think it would be good for it to exist, but I am not sure how a permanent frozen release fits into the terra versioning system. I am currently looking at [zed](https://github.com/terrapkg/packages/tree/frawhide/anda/devs/zed) as an example for handling stable/nightly versions of a package.

Is there a good/standard/suggested way to do a frozen/perma release version of a package?

_This is referencing #8537, and should be closed along with it_

arbormoss:
This adds tooling for [uxn](https://100r.co/site/uxn.html), the uxn half of the uxn/varvara computing stack.

This is a draft dependent on a rolling release version in addition to the current permanently frozen stable version.

See this [Discord message](https://discord.com/channels/957104883490775061/1209327267852058704/1452519075187261543) for the full versioning issue for conversation on the versioning issue.

The stable version is currently functional.

Implementing the nightly version is TODO at time of initial draft PR.

_This should close #8538_

This duplicates #8809 , feel free to use either, they are alternate implementations of the same package.

### DIFF
--- a/anda/misc/uxn/stable/uxn.spec
+++ b/anda/misc/uxn/stable/uxn.spec
@@ -1,0 +1,41 @@
+# uxn stable has no update script because this version
+# is permanently frozen.
+
+%define debug_package %nil
+
+Name:           uxn
+Version:        1.0
+Release:        1%?dist
+Summary:        An emulator for the Uxn computing stack
+URL:            https://100r.ca/site/%{name}.html
+Source0:        https://git.sr.ht/~rabbits/%{name}/archive/%{version}.tar.gz
+License:        MIT
+BuildRequires:  anda-srpm-macros SDL2-devel gcc
+Requires:       SDL2-devel
+
+Packager:       arbormoss <arbormoss@woodsprite.dev>
+
+%description
+%summary.
+
+%prep
+%autosetup -n %name-%version
+
+%build
+./build.sh
+
+%install
+install -Dm755 bin/%{name}asm %{buildroot}%{_bindir}/%{name}asm
+install -Dm755 bin/%{name}cli %{buildroot}%{_bindir}/%{name}cli
+install -Dm755 bin/%{name}emu %{buildroot}%{_bindir}/%{name}emu
+
+%files
+%doc README.md
+%license LICENSE
+%{_bindir}/%{name}asm
+%{_bindir}/%{name}cli
+%{_bindir}/%{name}emu
+
+%changelog
+* Sun Dec 21 2025 arbormoss <arbormoss@woodsprite.dev>
+- Intial Commit
--- a/anda/misc/uxn/stable/uxn.spec
+++ b/anda/misc/uxn/stable/uxn.spec
@@ -2,6 +2,8 @@
 # is permanently frozen. A nightly will be added once the
 # sourcehut versioning script is done.
 
+%global debug_package %{nil}
+
 Name:           uxn
 Version:        1.0
 Release:        1%?dist
@@ -20,7 +22,6 @@ Packager:       arbormoss <arbormoss@woodsprite.dev>
 %autosetup -n %name-%version
 
 %build
-sed -zi 's/debug=0/debug=1/' ./build.sh
 ./build.sh
 
 %install
--- a/anda/misc/uxn/stable/uxn.spec
+++ b/anda/misc/uxn/stable/uxn.spec
@@ -1,5 +1,6 @@
 # uxn stable has no update script because this version
-# is permanently frozen.
+# is permanently frozen. A nightly will be added once the
+# sourcehut versioning script is done.
 
 %define debug_package %nil
 
--- a/anda/misc/uxn/stable/uxn.spec
+++ b/anda/misc/uxn/stable/uxn.spec
@@ -2,8 +2,6 @@
 # is permanently frozen. A nightly will be added once the
 # sourcehut versioning script is done.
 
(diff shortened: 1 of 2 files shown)


arbormoss:
This PR is ready.

The files are currently in `uxn/stable` in anticipation of a nightly.

uxn nightly is dependent on a sourcehut nightly macro, which has some issues in #8538 .

This PR was tested on terra-43 because of issues right now with frawhide.

anda/misc/uxn/stable/uxn.spec
```diff
@@ -0,0 +1,42 @@
+# uxn stable has no update script because this version
+# is permanently frozen. A nightly will be added once the
+# sourcehut versioning script is done.
+
+%define debug_package %nil
```
metcya: Is this needed?

Looking around `build.sh`, you might be able to just patch `CFLAGS` to include the `-g` flag.

anda/misc/uxn/stable/uxn.spec
```diff
@@ -0,0 +1,42 @@
...
+
+Name:           uxn
+Version:        1.0
+Release:        1%?dist
+Summary:        An emulator for the Varvara virtual machine
+URL:            https://100r.ca/site/%{name}.html
+Source0:        https://git.sr.ht/~rabbits/%{name}/archive/%{version}.tar.gz
+License:        MIT
+BuildRequires:  anda-srpm-macros SDL2-devel gcc
+Requires:       SDL2-devel
```
metcya: Does this need the headers, etc from SDL2 at runtime, or just the shared object file? If not, you don't need `Requires: SDL2-devel`

metcya (changes requested):

anda/misc/uxn/stable/uxn.spec
```diff
@@ -0,0 +1,42 @@
...
+%define debug_package %nil
+
+Name:           uxn
+Version:        1.0
+Release:        1%?dist
+Summary:        An emulator for the Varvara virtual machine
+URL:            https://100r.ca/site/%{name}.html
+Source0:        https://git.sr.ht/~rabbits/%{name}/archive/%{version}.tar.gz
+License:        MIT
+BuildRequires:  anda-srpm-macros SDL2-devel gcc
```
metcya:
```suggestion
BuildRequires:  SDL2-devel gcc
```
No anda specific macros are being used here, so `anda-srpm-macros` isn't needed.

anda/misc/uxn/stable/uxn.spec
arbormoss: Patching -g directly doesn't work because if the internal debug variable isn't set to 1 it will make sure to not debug:
```sh
# ./build.sh
. . .
if [ $debug = 1 ];
then
	echo "[debug]"
	CFLAGS="${CFLAGS} -DDEBUG -Wpedantic -Wshadow -Wextra -Werror=implicit-int -Werror=incompatible-pointer-types -Werror=int-conversion -Wvla -g -Og -fsanitize=address -fsanitize=undefined"
else
	CFLAGS="${CFLAGS} -DNDEBUG -O2 -g0 -s"
fi
. . .
```

I just pushed a fix that just uses sed to set the debug var within `build.sh`.

metcya: In that case I think it would be better to keep the `%global debug_package %{nil}` than set the debug var. I think its better to not have the debug package than to distribute an unoptimized debug build of the package.

arbormoss: Change pushed to disable debug again.

Would it be better to just not use `build.sh` at all?

#8809 is an example of this, it rewrites the `build.sh` script, removing command line interactivity and using the `%{build_cflags}` macro.

This is all that you need to replicate the effect of the build script once you remove interactivity, formatting, etc.

```
CC="${CC:-cc}"
UXNEMU_LDFLAGS="-L/usr/local/lib $(sdl2-config --cflags --libs)"
CFLAGS="%{build_cflags}"

${CC} ${CFLAGS} src/%{name}asm.c -o bin/%{name}asm
${CC} ${CFLAGS} src/%{name}.c src/devices/system.c src/devices/console.c src/devices/file.c src/devices/datetime.c src/devices/mouse.c src/devices/controller.c src/devices/screen.c src/devices/audio.c src/%{name}emu.c ${UXNEMU_LDFLAGS} ${FILE_LDFLAGS} -o bin/%{name}emu
${CC} ${CFLAGS} src/%{name}.c src/devices/system.c src/devices/console.c src/devices/file.c src/devices/datetime.c src/%{name}cli.c ${FILE_LDFLAGS} -o bin/%{name}cli
```

It works fine with debug enabled.

arbormoss:
Feel free to use either this (#8537 ) or #8809 and close the other.

This uses the provided `build.sh`.

#8809 does not, and has debug properly enabled.

PR #8726, opened 2025-12-28
fix(vicinae): package vicinae-xdgpp

metcya:


Owen-sz:
Closed by https://github.com/terrapkg/packages/pull/9945

PR #9392, opened 2026-01-20
fix (picotool): version

Owen-sz:


Owen-sz:
Oh it's sanetized_ver time

PR #9467, opened 2026-01-24
fix(vpkedit): Exec key path

GildedRoach:


Owen-sz:
Superseded by https://github.com/terrapkg/packages/pull/10438

PR #9471, opened 2026-01-25
add: textual

Owen-sz:
Needs

```
  - nothing provides (python3.14dist(pygments) < 3~~ with python3.14dist(pygments) >= 2.19.2) needed by python3-textual-7.3.0-1.fc43.noarch from @commandline
  - nothing provides python3.14dist(rich) >= 14.2 needed by python3-textual-7.3.0-1.fc43.noarch from @commandline
```


metcya (changes requested):
not spec file enough

GildedRoach:
Bad news on the requires, IIRC Fedora has Pygments.

Owen-sz:
> Bad news on the requires, IIRC Fedora has Pygments.

Oh joy, will test patching that version req tonight. Looks like fedora also has rich...

PR #9621, opened 2026-02-03
twintaillauncher

solomoncyj:
TwinTail Launcher https://github.com/TwintailTeam/TwintailLauncher

**Describe the motivation**
Alternative to Bin-only Copr. Allows user to play Hoyo games

**Additional context**
N/A


anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,63 @@
...
+%description
+Twintaillauncher is a multi-platform launcher that brings mod support, quality-of-life improvements, and advanced features to a variety of anime-styled games.
+TTL is an all-in-one tool for downloading, managing, and launching your favorite anime games. It’s designed with flexibility, ease of use, and customization in mind.
+
+%prep
+%autosetup -n %{name}
+%tauri_prep
+
+%build
+%pnpm_build -v
```
GildedRoach: `-v` is for running Electron Vite build steps, is this what you wanted?

solomoncyj: yup, there is a vite config file in the pacakge root

GildedRoach: They use regular Vite and not Electron Vite (which is what I would expect since this isn't using Electron). They're doing similar things but Vite is usually part of other build scripts, Electron Vite is usually called on its own (hence the build flag option for it).

solomoncyj: oh TIL

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,63 @@
...
+
+
+%files
+%license LICENSE.dependencies
+%license LICENSE
+%doc README.md
+
+%{_bindir}/twintaillauncher
+%{_hicolordir}/128x128/apps/%{name}.png
+%{_hicolordir}/128x128/apps/%{name}.png
```
Owen-sz: Why is the same time listed twice here? Seems one should be 512x512

anda/apps/twintaillauncher/twintaillauncher.spec
GildedRoach: Electron is just Special :tm:.

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,63 @@
...
+%build
+%pnpm_build
+
+
+%install
+%tauri_install
+%tauri_cargo_license_summary
+%{tauri_cargo_license} > LICENSE.dependencies
+
+%desktop_file_install -f  ./twintaillauncher.desktop
```
Owen-sz, line 55: This isn't listed in %files

Owen-sz (changes requested):

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,70 @@
+
```
GildedRoach, line 2: Classic RPM trying to mangle Rust files moment.
```suggestion
%undefine __brp_mangle_shebangs
```

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,70 @@
...
+Name:           twintaillauncher
+%electronmeta
+
+Version:        1.1.15
+Release:        1%{?dist}
+Summary:        A multi-platform launcher for your anime games
+Packager:        Yoong Jin <solomoncyj@gmail.com>
+
+License:        GPL-3.0 AND %{electron_license}
```
Owen-sz: The GPL-3.0 needs to be either
`GPL-3.0-only`
`GPL-3.0-or-later`
Depending on what the license actually is.

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,70 @@
...
+%tauri_prep
+
+%build
+%pnpm_build
+
+
+%install
+%tauri_install
+%tauri_cargo_license_summary
+%{tauri_cargo_license} > LICENSE.dependencies
```
Owen-sz, line 53: You should put the licenses it puts in this file in License: tag as well (if there are any that are not already covered)

anda/apps/twintaillauncher/update.rhai
```diff
@@ -0,0 +1 @@
+rpm.version(gh_tag("TwintailTeam/TwintailLauncher"));
```
Owen-sz: Does this return `ttl-v1.1.15`? If it does, see what it returns without the `_tag`

If both return the `ttl-`, we should remove that, we can figure that out if needed.

(You can test easily by replacing `rpm.version` with `print`, and then running the file with `anda run` with a GitHub token

Owen-sz (changes requested):

anda/apps/twintaillauncher/twintaillauncher.spec
GildedRoach: This would also be the source license because, well, Cargo.

Electron license doesn't apply here.

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,70 @@
...
+install -Dm644 public/launcher-icon-128.png %{buildroot}%{_hicolordir}/128x128/apps/%{name}.png
+
+
+%files
+%license LICENSE.dependencies
+%license LICENSE
+%doc README.md
+
+%{_bindir}/twintaillauncher
+%{_hicolordir}/ 512x512/apps/%{name}.png
```
Owen-sz:
```suggestion
%{_hicolordir}/512x512/apps/%{name}.png
```

Owen-sz (changes requested):

tulilirockz:
Note that this is packaged as a flatpak https://flathub.org/en/apps/app.twintaillauncher.ttl

GildedRoach:
> Note that this is packaged as a flatpak https://flathub.org/en/apps/app.twintaillauncher.ttl

That's fine. We have an exception for game launchers.

GildedRoach:
The build failure on x86_64 should be fixed if you add `mock = 1` to the `anda.hcl`, I did a change to fix this CI side.

Is there a reason you forced this to build only on x86_64?

solomoncyj:
The game and the addition deps the app uses are x64 only

Owen-sz:
I don't see any x86_64 exclusive deps? But if they only want it on x86_64 that's fine.

Could you add that mock = 1 to the anda.hcl? I'd like to see this build, then there's a few minor things left, then we should be good

Owen-sz:
BTW it needs to be formatted like this

https://github.com/terrapkg/packages/blob/frawhide/anda/games/steam/anda.hcl

Owen-sz:
`BuildRequires: mold`

GildedRoach:
`%tauri_install` does the funny Fedora thing of installing the build files, this is identical to the behavior of `%cargo_install` and *may* require the kind of devel subpackages you see in Fedora Rust specs made using `rust2rpm`. In this case, it would.

If you want to install *just* the binary, we have `%tauri_install_bin` which is a wrapper around our `%crate_install_bin` which only installs a built binary.

solomoncyj:
Finally...

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,78 @@
+%undefine __brp_mangle_shebangs
+
+%global toolchain clang
+
+Name:           twintaillauncher
+%electronmeta
```
GildedRoach: You shouldn't mix Electron and Tauri macros.
```suggestion
```

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -0,0 +1,78 @@
...
+Packager:        Yoong Jin <solomoncyj@gmail.com>
+
+SourceLicense: GPL-3.0-or-later
+License:        GPL-3.0-or-later AND (((Apache-2.0 OR MIT) AND BSD-3-Clause) AND ((MIT OR Apache-2.0) AND Unicode-3.0) AND (0BSD OR Apache-2.0 OR MIT) AND (Apache-2.0) AND (Apache-2.0 AND ISC) AND (Apache-2.0 AND MIT) AND (Apache-2.0 OR Apache-2.0 WITH LLVM-exception OR CC0-1.0) AND (Apache-2.0 OR Apache-2.0 WITH LLVM-exception OR MIT) AND (Apache-2.0 OR BSD-2-Clause OR MIT) AND (Apache-2.0 OR BSD-3-Clause) AND (Apache-2.0 OR BSD-3-Clause OR MIT) AND (Apache-2.0 OR BSL-1.0 OR MIT) AND (Apache-2.0 OR CC0-1.0 OR MIT-0) AND (Apache-2.0 OR ISC OR MIT) AND (Apache-2.0 OR LGPL-2.1-or-later OR MIT) AND (Apache-2.0 OR MIT) AND (Apache-2.0 OR MIT OR Zlib) AND (Apache-2.0 WITH LLVM-exception) AND (BSD-2-Clause) AND (BSD-3-Clause) AND (BSD-3-Clause AND MIT) AND (BSD-3-Clause OR MIT) AND (CC0-1.0) AND (CC0-1.0 OR MIT-0) AND (CDLA-Permissive-2.0) AND (ISC) AND (ISC AND (Apache-2.0 OR ISC)) AND (ISC AND (Apache-2.0 OR ISC) AND OpenSSL) AND (LGPL-3.0-or-later OR MIT) AND (MIT) AND (MIT OR Unlicense) AND (MPL-2.0) AND (N/A) AND (Unicode-3.0) AND (Zlib) AND (bzip2-1.0.6))
```
GildedRoach: Single licenses shouldn't have `()`, ones in groups  (like `(LICENSE-1 OR LICENSE-2)`) are correct however.

GildedRoach (approved):

Owen-sz (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|el10|[](https://github.com/terrapkg/packages/pull/9684)|
|✅|f42|[](https://github.com/terrapkg/packages/pull/9685)|
|✅|f43|[](https://github.com/terrapkg/packages/pull/9686)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/21701499759) for details

PR #9626, opened 2026-02-03
osu-lazer.spec: use EOF in quotes to escape shell variables

Opposite34:
So for osu-lazer, the spec file is broken and it looked like this previously:
```
#!/bin/sh
env OSU_EXTERNAL_UPDATE_PROVIDER=1 /opt/osu-lazer/osu.AppImage ""
```

but it should be `env OSU_EXTERNAL_UPDATE_PROVIDER=1 /opt/osu-lazer/osu.AppImage "$@"` to accept files.

This is because `EOF` in cat needs to be quoted (`'EOF'`) to prevent variable expansion. See https://stackoverflow.com/a/27921346

### DIFF
--- a/anda/games/osu-lazer/osu-lazer.spec
+++ b/anda/games/osu-lazer/osu-lazer.spec
@@ -4,7 +4,7 @@
 
 Name:			osu-lazer
 Version:		2026.119.0
-Release:		1%?dist
+Release:		2%?dist
 Summary:		The future of osu! and the beginning of an open era! Commonly known by the codename osu!lazer. Pew pew.
 ExclusiveArch:	x86_64
 URL:			https://osu.ppy.sh/
--- a/anda/games/osu-lazer/osu-lazer.spec
+++ b/anda/games/osu-lazer/osu-lazer.spec
@@ -21,7 +21,7 @@ Source4:		osu-lazer-uri-handler.desktop
 %{summary}
 
 %prep
-cat <<EOF > osu-lazer
+cat <<'EOF' > osu-lazer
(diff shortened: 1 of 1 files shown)


Owen-sz:
Thanks for the PR!
One quick thing, could you bump the '1' in the `Release:` tag to a '2'?

Opposite34:
Note that this makes `osu-lazer.desktop` not appear in open with anymore. However, you *can* use the uri handler to handle the files as well. If this is not ideal I think putting `%F` for the args (like so: `Exec=/usr/bin/osu-lazer %F`) in the osu-lazer desktop file is a proper way to handle this.

Opposite34:
> Thanks for the PR! One quick thing, could you bump the '1' in the `Release:` tag to a '2'?

bumped! Also can you look at the desktop file if the `%F` is preferred as well in the desktop file? @Owen-sz 
(edited: seems like the `%F` is not needed upon building and testing the package so this is fine)

Owen-sz (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|el10|[](https://github.com/terrapkg/packages/pull/9627)|
|✅|f42|[](https://github.com/terrapkg/packages/pull/9628)|
|✅|f43|[](https://github.com/terrapkg/packages/pull/9629)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/21641875667) for details

PR #9690, opened 2026-02-06
ogc-wireplumber: init package

tulilirockz:
This is just the wireplumber with the OGC patches, needs to be manually verified whenever we add new patches or not (hopefully not). This current one just has a fix for wpctl on gamescope-session.

DRAFT: Not 100% sure this package is valid yet.


Owen-sz:
Can you add a changelog entry 'Port to Terra from xxx'

PR #9826, opened 2026-02-12
fix(vicinae): bdep Qt6GuiPrivate

madonuko:


Owen-sz:
Related PR https://github.com/terrapkg/packages/pull/8726

PR #9973, opened 2026-02-20
add: gf

metcya:


metcya:
cucked by gcc-16,,,

Owen-sz:
Does this build on <= 44?
We could pull in different GCC versions

GildedRoach:
> Does this build on <= 44? We could pull in different GCC versions

Does Rawhide have `gcc-15` yet?

metcya:
i haven't tested on 44, but it builds on 43 with gcc 15

metcya:
tbh this should probably be patched though. I *think* this is UB.

metcya:
i'm just that goated

GildedRoach:
Does this still build on GCC15?

Owen-sz:
Goated

Owen-sz:
nightly label in `anda.hcl`?

anda/tools/gf/gf.spec
```diff
@@ -0,0 +1,48 @@
+%global forgeurl https://github.com/nakst/gf
+
+%global commit 9a5dbcc90dc9ca9580f6ce2854cd67e2e507b0c1
+%global shortcommit %{sub %{commit} 0 7}
+%global commitdate 20251231
+
+%forgemeta
+
+Name:           gf
+Version:        0^%{commitdate}.%{shortcommit}
```
Owen-sz:
```suggestion
Version:        0~%{commitdate}git.%{shortcommit}
```

Owen-sz (changes requested):

anda/tools/gf/gf.spec
metcya: fedora requires `^` for snapshots. see the [fedora packaging guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_snapshots) on this.

`^` always sorts higher, which is the behavior we want for snapshots. `~` sorts lower, which is the behavior needed for prereleases/release candidates.

for example
```
1.2.3^20260220.abcdef > 1.2.3 > 1.2.3~rc1
```

GildedRoach: I agree with you, this versioning came about due to a disagreement among our packagers. I am on the side we should follow the Fedora guidelines on this.

However, `git` is technically more correct here than `.`.

metcya:
> Does this still build on GCC15?

yes, it still compiles fine on gcc-15. mixing ordinary and designated initializers is valid in c, but invalid in c++. gcc-15 doesn't issue a diagnostic for this, but gcc-16 does.

GildedRoach:
> > Does this still build on GCC15?
> 
> yes, it still compiles fine on gcc-15. mixing ordinary and designated initializers is valid in c, but invalid in c++. gcc-15 doesn't issue a diagnostic for this, but gcc-16 does.

You can do `BuildRequires: gcc15 gcc15-c++` then `%global __cc gcc-15` and `%global __cxx g++-15` if Rawhide has these already to avoid a patch, otherwise the patch might be fine.

metcya:
> > > Does this still build on GCC15?
> > 
> > 
> > yes, it still compiles fine on gcc-15. mixing ordinary and designated initializers is valid in c, but invalid in c++. gcc-15 doesn't issue a diagnostic for this, but gcc-16 does.
> 
> You can do `BuildRequires: gcc15 gcc15-c++` then `%global __cc gcc-15` and `%global __cxx g++-15` if Rawhide has these already to avoid a patch, otherwise the patch might be fine.

I think keeping the patch is fine. gcc-15 _should_ issue a diagnostic here, and _technically_ the program is using undefined behavior without the patch.

anda/tools/gf/gf.spec
Owen-sz: I also agree but currently our guidelines say otherwise...

Owen-sz (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|el10|[](https://github.com/terrapkg/packages/pull/10009)|
|✅|f42|[](https://github.com/terrapkg/packages/pull/10010)|
|✅|f43|[](https://github.com/terrapkg/packages/pull/10011)|
|✅|f44|[](https://github.com/terrapkg/packages/pull/10012)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/22231813305) for details

PR #9997, opened 2026-02-20
feat: GPG update script to auto fetch new keys and emergency force update

GildedRoach:
"WTF is this?" you may rightfully ask.

The answer:
- Update script that fetches new GPG keys only if manual ones do not exist for a given repo (for server ping, CI time, and security reasons).
- Changed to `rpm.version()` function (though it's very hacky) to handle `rpm.release()` better for key changes.
- Emergency "oh shit" button workflow in the event our keys are ever compromised that should force overwrite them. IT IS STILL PREFERABLE TO MANUALLY DO THIS but this is an emergency "oh God no one is at their computer" workflow.

All of this is automagic as long as we update other workflows.


madonuko:
why not put all the curling into pre.rhai?

GildedRoach:
> why not put all the curling into pre.rhai?

Because 5,000 `sh()`s basically and I wasn't sure if I'd need to escape anything with how I checked if the files existed.

I didn't use `get()` because Rhai conditionals are way more massive than what I used.

GildedRoach:
> > why not put all the curling into pre.rhai?
> 
> Because 5,000 `sh()`s basically and I wasn't sure if I'd need to escape anything with how I checked if the files existed.



> > why not put all the curling into pre.rhai?
> 
> Because 5,000 `sh()`s basically and I wasn't sure if I'd need to escape anything with how I checked if the files existed.
> 
> I didn't use `get()` because Rhai conditionals are way more massive than what I used.

Actually, correction, zero way to bump the release on the spec if a new key was added this way. I misread this as `update.rhai` at first.

madonuko:
well this is the kind of package you want to build in bootstrap.yml, so not relying on the update.rhai logic should be better

GildedRoach:
> well this is the kind of package you want to build in bootstrap.yml, so not relying on the update.rhai logic should be better

I'm confused? This is *just* to fetch key files which don't already exist (to avoid the issue we just had where no one had added the 44 keys), it can still be built in `bootstrap.yml` and I also planned to add it to that.

madonuko (approved):

halfcyan:
update on this?

Issue #10155, opened 2026-02-28
[BUG] kde-material-you-colors is not working and outdated


veuxit:
### Full Raw Package Name (e.g. anda-0.4.14-1.fcrawhide.x86_64)

kde-material-you-colors-2.0.0-5.fc43.x86_64

### Release Version

f43

### Describe the Bug

Not working and outdated (2.0.2 is available)
Can be updated manually, automatic update script is broken 

### To Reproduce

1. dnf install kde-material-you-colors
2. kde-material-you-colors
3. error

### Expected Behavior

It should work??

### Relevant Log Output

```shell
Traceback (most recent call last):
  File "/home/veux/.local/bin/kde-material-you-colors", line 5, in <module>
    from kde_material_you_colors.main import main
  File "/usr/lib/python3.14/site-packages/kde_material_you_colors/main.py", line 552, in <module>
    main()
    ~~~~^^
  File "/usr/lib/python3.14/site-packages/kde_material_you_colors/main.py", line 379, in main
    from . import apply_themes
  File "/usr/lib/python3.14/site-packages/kde_material_you_colors/apply_themes.py", line 5, in <module>
    from kde_material_you_colors.utils import (
    ...<8 lines>...
    )
  File "/usr/lib/python3.14/site-packages/kde_material_you_colors/utils/m3_scheme_utils.py", line 29, in <module>
    from kde_material_you_colors.utils.extra_image_utils import sourceColorsFromImage
  File "/usr/lib/python3.14/site-packages/kde_material_you_colors/utils/extra_image_utils.py", line 10, in <module>
    SCORE_OPTIONS = ScoreOptions(
        desired=7,  # ANSI colors target
    ...<2 lines>...
        dislike_filter=True,  # Fix globally disliked colors
    )
TypeError: ScoreOptions.__init__() got an unexpected keyword argument 'dislike_filter'
```

veuxit:
Seems package is broken even if installed from developers repo

madonuko:
@hamachitan kde-material-you-colors-2.0.0-5.fc43.x86_64

madonuko:
> Seems package is broken even if installed from developers repo

interesting, yeah I think there are some version mismatches in python deps…

veuxit:
Well seems to be fixed 
Still got this error but seems to be source issue and package is working 
```[E] konsole_utils: reload_profile: org.freedesktop.DBus.Error.AccessDenied: Security sensitive DBus API is disabled in the settings.
[...]

PR #10156, opened 2026-02-28
kde-material-you-colors: bump to 2.0.2 and apply temp fix

veuxit:
This fixes this https://github.com/terrapkg/packages/issues/10155

### DIFF
--- a/anda/themes/kde-material-you-colors/kde-material-you-colors.spec
+++ b/anda/themes/kde-material-you-colors/kde-material-you-colors.spec
@@ -3,7 +3,7 @@
 %global org "com.github.luisbocanegra"
 
 Name:           kde-material-you-colors
-Version:        2.0.0
+Version:        2.0.2
 Release:        4%?dist
 Summary:        Automatic Material You Colors Generator from your wallpaper for the Plasma Desktop
 License:        GPL-3.0-only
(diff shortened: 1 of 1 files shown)


GildedRoach:
The change to `python3dist(materialyoucolor)` will not work as Terra only carries one version of a package at a time, this swaps a runtime issue for a package dependency one.

veuxit:
> The change to `python3dist(materialyoucolor)` will not work as Terra only carries one version of a package at a time, this swaps a runtime issue for a package dependency one.

What to do then? However, I built this locally and it was normal with terra-44-x86_64

GildedRoach:
> > The change to `python3dist(materialyoucolor)` will not work as Terra only carries one version of a package at a time, this swaps a runtime issue for a package dependency one.
> 
> What to do then? However, I built this locally and it was normal with terra-44-x86_64

Does it still work if you don't change the `materialyoucolor` dep?

veuxit:
> Does it still work if you don't change the `materialyoucolor` dep?

No, there are breaking changes in materialyoucolor dep since version 2.0.10
https://github.com/luisbocanegra/kde-material-you-colors/issues/304#issuecomment-3973682630

GildedRoach:
> 
> > Does it still work if you don't change the `materialyoucolor` dep?
> 
> No, there are breaking changes in materialyoucolor dep since version 2.0.10
> https://github.com/luisbocanegra/kde-material-you-colors/issues/304#issuecomment-3973682630

Well, here's the issue. This is on 43. 

As you can see, only versions above that currently exist.

veuxit:
We need better solution or just wait for developer to fix this

PR #10382, opened 2026-03-08
chore: update gsettings-qt-lomiri to have an update script, update version, fix spec

Owen-sz:


anda/desktops/lomiri-unity/gsettings-qt-lomiri/update.rhai
```diff
@@ -0,0 +1 @@
+print(gitlab_tag("10070038"));
```
madonuko: rpm.version?

PR #10403, opened 2026-03-10
fix: vicinae

Owen-sz:


GildedRoach:
Optflags is the shared C* flags (as in what C and C++ share, though technically it has all thre C flags IIRC just not C++), you'll want to call either `%{build_cxxflags}` or `$CXXFLAGS`.

PR #10489, opened 2026-03-13
fix: bump gamescope-session(and -opengamepadui) for f44

porkloin:
Last two F44 packages we need that are busted on [zirconium-jackrabbit](https://github.com/zirconium-dev/zirconium-jackrabbit)


Owen-sz:
Thanks!

Owen-sz (approved):

PR #10678, opened 2026-03-16
chore (arduino-app-lab): build from source, rename to remove the -bin, fix licenses, clean up spec

Owen-sz:


Owen-sz:
Of course this doesn't work in CI because why would it

madonuko (dismissed):

PR #10688, opened 2026-03-17
fix(prismlauncher): rename modrinth-mrpack-mime.xml to org.prismlauncher.PrismLauncher.xml

idiot-shinuki:
….xml

PrismLauncher/PrismLauncher#5202 renames a build artifact. This commit reflects that change in the .spec file for prismlauncher-nightly, which is the only package where the change is present as of this commit.


idiot-shinuki:
Might wanna wait for this, see PrismLauncher/PrismLauncher#5205

Owen-sz:
Rawhide is still misbehaving. I'll test this locally on 43 in a few hours.

madonuko:
I'm getting a file not found on that file, I looked at the cmake logs and it seems to be the case that it's still installed to `/usr/share/mime/packages/modrinth-mrpack-mime.xml`

idiot-shinuki:
> I'm getting a file not found on that file, I looked at the cmake logs and it seems to be the case that it's still installed to `/usr/share/mime/packages/modrinth-mrpack-mime.xml`

Interesting, I wonder if they did something to revert the change? I'll try building the latest commit on my pc with the updated file name and see if it works.

madonuko (approved):

PR #11050, opened 2026-04-04
add: logitech-rs50-linux-driver

LuanVSO:
add packages for wheel driver logitech-rs50-linux-driver


LuanVSO:
don't know what the error in the ci is about, running the command locally on f43 works just fine:
image

PR #11067, opened 2026-04-04
Terra twintaillauncher promoted to official package & 2.0.0 bump

TukanDev:
This pull request promotes terra package of my project to officially adopted distribution source and bumps it to latest 2.0.0 release.

I do not know if i should handle anything else so please do inform me!


anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -1,86 +1,96 @@
...
-
-%global toolchain clang
-
 Name:           twintaillauncher
-
 Version:        2.0.0
 Release:        1%{?dist}
 Summary:        A multi-platform launcher for your anime games
-Packager:        Yoong Jin <solomoncyj@gmail.com>
+Packager:       TukanDev <contact@tukandev.com>
```
korewaChino, line 5: removal of attribution!?

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -1,86 +1,96 @@
...
+License:        GPL-3.0-only
 URL:            https://twintaillauncher.app/
-Source0:        https://github.com/TwintailTeam/TwintailLauncher/archive/refs/tags/ttl-v%{version}.tar.gz
+Source0:        https://github.com/TwintailTeam/TwintailLauncher/releases/download/ttl-v%{version}/twintaillauncher_%{version}_amd64.deb
```
korewaChino, line 9: DEB binary source should NOT be used unless upstream is closed source, in this case it is not

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -1,86 +1,96 @@
...
+cp -p usr/share/applications/twintaillauncher.desktop %{buildroot}%{_datadir}/applications/
+cp -p usr/share/icons/hicolor/32x32/apps/twintaillauncher.png \
+   %{buildroot}%{_datadir}/icons/hicolor/32x32/apps/
+cp -p usr/share/icons/hicolor/128x128/apps/twintaillauncher.png \
+   %{buildroot}%{_datadir}/icons/hicolor/128x128/apps/
+cp -p usr/share/icons/hicolor/256x256@2/apps/twintaillauncher.png \
+   %{buildroot}%{_datadir}/icons/hicolor/256x256@2/apps/
+cp -p usr/lib/twintaillauncher/resources/winetricks %{buildroot}/usr/lib/twintaillauncher/resources/
+cp -p usr/lib/twintaillauncher/resources/reaper %{buildroot}/usr/lib/twintaillauncher/resources/
+cp -p usr/lib/twintaillauncher/resources/hkrpg_patch.dll %{buildroot}/usr/lib/twintaillauncher/resources/
```
korewaChino, line 64: ditto.

anda/apps/twintaillauncher/twintaillauncher.spec
```diff
@@ -1,86 +1,96 @@
...
-%{_hicolordir}/512x512/apps/%{name}.png
-%{_hicolordir}/128x128/apps/%{name}.png
-%_appsdir/twintaillauncher.desktop
-
-
-
+%{_datadir}/applications/twintaillauncher.desktop
+%{_datadir}/icons/hicolor/32x32/apps/twintaillauncher.png
+%{_datadir}/icons/hicolor/128x128/apps/twintaillauncher.png
+%{_datadir}/icons/hicolor/256x256@2/apps/twintaillauncher.png
```
korewaChino, line 88: just wildcard the icon sizes

korewaChino (changes requested):

TukanDev:
Just to put little note I will be able to resolve requested changes in a week as I am on vacation unless some of you kind reviewers want to fix these simple ones like license and images wildcard as they are very simple edits license line can be copied from old spec file.

As for deb I am using deb as it matches same spec file used for our copr repository so I would prefer for them to be as identical as possible but if not possible i will deal with it once I get back.

Owen-sz:
We can not merge this PR in its current state due to:
- We are not changing it from being built from source, to being fetched from a Deb of all things. I see you’re pushing this method as the preferred distribution method on your README, this should not be done.
- Licenses being removed so this is no longer properly licensed for an RPM
- The %post, %postun, and %posttrans scripts are all unnecessary
- Removing the original packager

None of this is acceptable and the last one is just rude. The only thing that looks good here is removing the %globals and %undefine, assuming that’s no longer needed here.
Also, the `Version:` was updated here - https://github.com/terrapkg/packages/commit/e154967dd96ca581ecee9765bdec600ed1f31bfa

anda/apps/twintaillauncher/twintaillauncher.spec
solomoncyj: authorized as they are trying to replace the spec file

PR #11506, opened 2026-04-18
fix (libfprint-tod): license identifier, improper %dir, use %conf, add packager line, missing bdeps

Owen-sz:


anda/lib/libfprint-tod/libfprint-tod.spec
```diff
@@ -2,11 +2,11 @@
 
 Name:           libfprint-tod
 Version:        1.95.1+tod1
-Release:        1%{?dist}
+Release:        2%{?dist}
 URL:            https://gitlab.freedesktop.org/3v1n0/libfprint/
 Source:         %{url}/-/archive/v%{version}/libfprint-v%{version}.tar.gz
 Summary:        a light fork of libfprint to expose internal Drivers API in order to create drivers as shared libraries
-License:        GPL-2.1
+License:        GPL-2.1-or-later
```
GildedRoach: This was actually just incorrect the whole time...
```suggestion
License:        LGPL-2.1-or-later
```

GildedRoach:
That build failure wasn't a build dependencies issue, and also this whole package is broken because it needs an SELinux policy and SELinux scriptlets to work on Fedora at all.

madonuko:
rename the spec?

PR #11654, opened 2026-04-26
chore: Sync f43

GildedRoach:


anda/devs/tracy/tracy.spec
```diff
@@ -6,7 +6,7 @@
 
 Name:			tracy
 Version:		0.13.1
-Release:		4%?dist
+Release:		3%?dist
```
Owen-sz: Pretty sure this puts Tracy out of sync

Owen-sz (changes requested):

anda/devs/tracy/tracy.spec
GildedRoach: Yeah it was never fixed on Rawhide... Hold on.

Owen-sz (approved):

PR #11790, opened 2026-04-30
add: rsms-inter-fonts

### LINKED ISSUE #10773: [Request] Inter Font for EL
rsms-inter-fonts is available for fedora and not for EL

Could you ship this for EL using terra?

https://github.com/rsms/inter

Owen-sz:
Only needed on el10, in rawhide for updates until we can move to packagesel repo

Resolves https://github.com/terrapkg/packages/issues/10773


GildedRoach:
You can add this only in EL and make it updranch.

Owen-sz:
> You can add this only in EL and make it updranch.

Won't it not update then?

GildedRoach:
> > You can add this only in EL and make it updranch.
> 
> Won't it not update then?

If you make it updbranch it will. Because `updbranch` means it runs in the per branch updates.

You can also do the funny hack and make it track Fedora 43 or 42 for version updates. (I suggest these because EL usually has older deps.)

PR #11835, opened 2026-05-03
add: TgWsProxy

veuxit:
The executable size is around 100M if i build this locally, but if i remove --system-site-packages flag then package size is ok but it doesnt work
Shipping this as is now, at least it works

### DIFF
--- a/anda/tools/tg-ws-proxy/update.rhai
+++ b/anda/tools/tg-ws-proxy/update.rhai
@@ -1,0 +1,1 @@
+rpm.version(gh_tag(Flowseal/tg-ws-proxy));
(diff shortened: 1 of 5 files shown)


veuxit:
Well size fixed but cant test it from actions since i am not on f45

anda/tools/tg-ws-proxy/update.rhai
```diff
@@ -0,0 +1 @@
+rpm.version(gh_tag(Flowseal/tg-ws-proxy));
```
madonuko:
```suggestion
rpm.version(gh_tag("Flowseal/tg-ws-proxy"));
```

madonuko:
> The executable size is around 100M if i build this locally, but if i remove --system-site-packages flag then package size is ok but it doesnt work

yeah `--system-site-packages` probably means it requires the pypi packages that are not bundled inside the rpm package, you will need to specify the dependency list.

GildedRoach:
This is not how system Python packages should be packaged. Please look at examples of our other packages using Fedora's Python macros, and yes you will need to package any runtime dependencies of this that aren't currently in Terra or Fedora.

PR #12645, opened 2026-05-26
fix: ghostty-nightly

Owen-sz:


GildedRoach:
Owen I've explained why this is wrong before :sob:

Owen-sz:
Wait lol where

GildedRoach:
> Wait lol where

I think on Signal but the tarball called `-source` unpacks into a folder name based on the shortcommit the version was released from.

Issue #12971, opened 2026-06-08
[Request] terra-curl


korewaChino:
**What software are you requesting to be packaged?**
The name of the software and link to website (if it exists) and source code (if available).

**Describe the motivation**
For some reason, Fedora's cURL package explicitly disables Zstd compression

https://src.fedoraproject.org/rpms/curl/blob/rawhide/f/curl.spec#_298

Enabling Zstd compression would be ideal for cURL for Zstd-compressed requests

**Additional context**
Add any other context about the package request here.

Owen-sz:
This has been fixed upstream https://src.fedoraproject.org/rpms/curl/c/9f1f3fca78c48dae018fa9bf19228903887564aa?branch=rawhide

PR #13018, opened 2026-06-09
add: terra-curl to extras

halfcyan:
as mentioned in #12971, fedora decided to remove zstd on-the-fly support from curl (?????) and this resolves it.


Owen-sz:
This has been fixed upstream https://src.fedoraproject.org/rpms/curl/c/9f1f3fca78c48dae018fa9bf19228903887564aa?branch=rawhide

PR #13168, opened 2026-06-17
chore: Sync el10

GildedRoach:


anda/system/asusctl/asusctl.spec
```diff
@@ -4,8 +4,8 @@
 %global asus_system_units asusd.service asus-shutdown.service
 
 Name:           asusctl
-Version:        6.3.8
-Release:        3%{?dist}
+Version:        1.0.1
```
Owen-sz, line 7: What happened here

Owen-sz: Oh the package bumped badly I'll fix that

Owen-sz (approved):

PR #13234, opened 2026-06-21
Add fonts suffix to Source Han subpackages

latin-1:


anda/fonts/source-han-sans/source-han-sans-fonts.spec
```diff
@@ -39,7 +39,7 @@ https://github.com/adobe-fonts/source-han-sans/blob/release/SourceHanSansReadMe.
 }%{-s:Subset: The fonts are split into regional-specific subset fonts.                              \
 }                                                                                                   \
                                                                                                     \
-%files %{-h:hw-}%{-v:vf-}%{-s:subset-}%1                                                            \
+%files %{-h:hw-}%{-v:vf-}%{-s:subset-}%1-fonts                                                      \
```
madonuko: this results in source-han-sans-fonts-hw-vf-subset-fonts for example, so maybe you should change this
```suggestion
%files -n source-han-sans-%{-h:hw-}%{-v:vf-}%{-s:subset-}%1-fonts                                                      \
```
you probably need -n for the other sections too

madonuko (approved):

PR #13828, opened 2026-07-09
zed: add missing terra-appstream-helper BuildRequires for %terra_appstream macro

cjatherton:
The zed.spec files are missing the `BuildRequires` for `terra-appstream-helper` and `python3-terra-appstream-helper` which provide the %terra_appstream macro.  Build fails without this at my COPR at https://copr.fedorainfracloud.org/coprs/cjatherton/zed/ .


Owen-sz:
Hey thanks for the PR!
These are actually not needed in the Terra spec specifically since they're pulled into our builder containers.

https://github.com/terrapkg/builder/blob/frawhide/Dockerfile

PR #13841, opened 2026-07-09
fix(envision): add -fPIE

madonuko:


GildedRoach:
Build error is weird but this is already set via the RH hardening.
image

GildedRoach:
Just noticed this issue, but the correct macro is `_pkg_extra_cflags` etc.

julian45:
> Build error is weird but this is already set via the RH hardening. image

Aha, the `fPIE` bit gave me a good hint :)
I just ran into a similar issue with Continuwuity earlier this week. I checked the action output, and [lo and behold](https://github.com/terrapkg/packages/actions/runs/29005052751/job/86074646478?pr=13841#step:8:1478): even though upstream doesn't explicitly call for it in [their Cargo.toml](https://gitlab.com/gabmus/envision/-/blob/main/Cargo.toml?ref_type=heads), they have a transitive dependency on aws-lc-rs and aws-lc-sys somewhere that must have been bumped to 0.42.0 recently. This is a known issue [super-upstream](https://github.com/aws/aws-lc-rs/issues/1168), and the [only known good workaround](https://forgejo.ellis.link/continuwuation/continuwuity/pulls/1908) is to explicitly pin these prior versions in the software being packaged: aws-lc-rs 1.17.0 and aws-lc-sys 0.41.0.

Ideally, the upstream software you're packaging should implement this themselves, but you might need to monkey-patch it here if they can't/won't.

julian45:
Supposedly, the aforementioned issue is fixed in aws-lc-rs 1.17.3/aws-lc-sys 0.43.0.

madonuko:
by the way _pkg_extra_flags should not be used, it's not supported on older releasevers

GildedRoach:
> by the way _pkg_extra_flags should not be used, it's not supported on older releasevers 

Source? Because everything in the Fedora information still mentions it and it's still in `redhat-rpm-config` on Rawhide.

julian45:
This workaround looks decent enough, but again, upstream bumping their aws-lc-rs/aws-lc-sys deps should resolve the need for a workaround in the first place. I opened [an issue upstream](https://gitlab.com/gabmus/envision/-/work_items/275) in hopes they will address this.

Owen-sz:
Seems like it has been fixed upstream, all that's needed here now is a rebuild

PR #13842, opened 2026-07-09
add: fastfetch

madonuko:
would like to invoke the new internal policy "broken packages in fedora can be conflicted in main terra repo"


Owen-sz:
Should only be merged if https://bugzilla.redhat.com/show_bug.cgi?id=2498476 is not fixed

PR #14294, opened 2026-07-22
chore: remove update scripts that fetch archive.ubuntu.com

madonuko:
These scripts are slowing down `anda update` to the point that on average it takes 10 minutes to finish.


Owen-sz:
Is there a better way to update these? Maybe a GitHub mirror or something?

PR #14315, opened 2026-07-23
fix: Don't point extra repos to Rawhide

GildedRoach:


Owen-sz:
Oh yeah forgor this, I'll add to the outline

Owen-sz (approved):

PR #14442, opened 2026-07-25
add: opendeck

NichSchlagen:
[OpenDeck](https://github.com/nekename/OpenDeck) is a desktop application for using stream controller devices like the Elgato Stream Deck. It supports plugins written against the original Stream Deck SDK and the OpenAction API.

Packaged at v2.13.1, GPL-3.0-or-later.

## Build

Verified with `anda build -c terra-rawhide-x86_64 anda/apps/opendeck/pkg` in `ghcr.io/terrapkg/builder:frawhide` — builds clean in ~9 minutes.

The frontend is built with **Deno**, not npm (`beforeBuildCommand` is `deno task build`), so this uses `%deno_build` rather than `%npm_build`. As far as I can tell this is the first spec in the repo to use that macro; it worked without any coaxing.

## Notes for review

**`openssl3-devel`.** Rawhide has moved to OpenSSL 4.0, and `openssl-sys` 0.9.111 refuses to build against it:

```
This crate is only compatible with OpenSSL (version 1.0.2 through 1.1.1, or 3),
or LibreSSL 3.5 through 4.2.x, but a different version of OpenSSL was found.
```

Only `reqwest` pulls it in, via `native-tls`. `openssl3-devel` provides `openssl-devel = 3.5.7` and conflicts with `>= 1:4.0`, so a plain `BuildRequires` is enough to displace the 4.0 one that `%tauri_buildrequires` would otherwise pick. The resulting package depends on `libssl.so.3` from `openssl3-libs`.

That compat package is marked `deprecated()`, so this is explicitly a stopgap and the spec says so. The alternative is patching `reqwest` onto `rustls-tls` — rustls is already in the dependency tree, so it would likely just work and would drop the openssl dependency entirely. I went with the compat lib because it leaves upstream's TLS stack alone, but I'm happy to switch if you'd prefer not to depend on a deprecated package.

**Resources in `/usr/lib`, not `%{_libdir}`.** Tauri resolves its resource directory relative to the binary, i.e. `/usr/bin/../lib/opendeck`, so the bundled starter pack plugin has to go to `/usr/lib/opendeck/plugins` even on 64-bit. Upstream's own RPM does the same.

**Plugins are installed by hand.** `src-tauri/build.rs` builds the starter pack into `src-tauri/target/plugins/`, and normally the Tauri bundler copies it into place — but `%tauri_common_opts` passes `--no-bundle`, so the spec copies it itself. I drop the `.crates.toml` / `.crates2.json` bookkeeping files that `cargo install` leaves behind; upstream's RPM ships them.

**Desktop file is ours.** Upstream only has a Tauri template with `{{placeholder}}`s that gets filled in during bundling, so the spec carries a real one as `Source1`. Compared to what upstream's bundler generates, I set `Categories=Utility;` instead of `Office;` (generated from their `"category": "Productivity"`, which seems wrong for this), and added `StartupWMClass` plus the `MimeType` scheme handlers that the deep-link plugin needs.

**udev rules** go to `%{_udevrulesdir}` rather than upstream's `/etc/udev/rules.d`, matching `8bitdo-udev-rules` and `wooting-udev-rules`.

The resulting file list matches upstream's RPM, and `%check` validates the desktop file.

### DIFF
--- a/anda/apps/opendeck/anda.hcl
+++ b/anda/apps/opendeck/anda.hcl
@@ -1,0 +1,5 @@
+project pkg {
+    rpm {
+        spec = "opendeck.spec"
+    }
+}
--- a/anda/apps/opendeck/opendeck.desktop
+++ b/anda/apps/opendeck/opendeck.desktop
@@ -1,0 +1,10 @@
+[Desktop Entry]
+Type=Application
+Name=OpenDeck
+Comment=Use stream controllers
+Exec=opendeck
+Icon=opendeck
+Terminal=false
+Categories=Utility;
+StartupWMClass=opendeck
+MimeType=x-scheme-handler/opendeck;x-scheme-handler/openaction;x-scheme-handler/streamdeck;
--- a/anda/apps/opendeck/opendeck.spec
+++ b/anda/apps/opendeck/opendeck.spec
@@ -1,0 +1,108 @@
+%undefine __brp_mangle_shebangs
+
+%global appid               opendeck
+%global name_pretty         OpenDeck
+%global appstream_component desktop-application
+%global developer           Aman Khanna
+# Tauri resolves its resource directory relative to the binary, i.e. /usr/bin/../lib/opendeck.
+# This is /usr/lib even on 64-bit, so %%_libdir must not be used here.
+%global opendeck_libdir     %{_prefix}/lib/%{name}
+
+Name:           opendeck
+Version:        2.13.1
+Release:        1%{?dist}
+Summary:        Use stream controllers
+URL:            https://github.com/nekename/OpenDeck
+Source0:        %{url}/archive/refs/tags/v%{version}.tar.gz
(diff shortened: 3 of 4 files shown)


madonuko (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|el10|[](https://github.com/terrapkg/packages/pull/14445)|
|✅|f43|[](https://github.com/terrapkg/packages/pull/14446)|
|✅|f44|[](https://github.com/terrapkg/packages/pull/14447)|
|✅|f45|[](https://github.com/terrapkg/packages/pull/14448)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/30200179469) for details

PR #14495, opened 2026-07-27
add: ghc-chs-deps

Owen-sz:

PR #14496, opened 2026-07-27
add: ghc-chs-cabal

Owen-sz:


Owen-sz:
Requires https://github.com/terrapkg/packages/pull/14495

Owen-sz:
So Fedora's cabal is too old for this...

PR #14713, opened 2026-07-30
add: ventoy

### LINKED ISSUE #3790: [Request] ventoy
ventoy
https://github.com/ventoy/Ventoy

**Describe the motivation**
Ventoy is a very popular tool to create a bootable USB with multiple ISOs, removing the need to format a USB drive over and over again. It isn't available as an RPM.

**Additional context**
There is a Linux GUI both in GTK and in QT (https://www.ventoy.net/en/doc_linux_gui.html), maybe this should be taken into consideration when packaging.

c4i0b:

**ventoy** — an open source tool to create bootable USB drives for ISO/WIM/IMG/VHD(x)/EFI files.

- Source: https://github.com/ventoy/Ventoy
- Website: https://www.ventoy.net
- License: GPL-3.0-or-later

**Describe the motivation**

Closes #3790. 

Ventoy is extremely popular (~78k★) and is **not available as an RPM** in Fedora or Terra. This gives Fedora users a clean, `dnf`-installable package with the GUI (`ventoygui` + `.desktop`, so it shows in the application menu and opens with a click, no terminal).

**Additional context**

- **Repackages the upstream prebuilt Linux release** (`ventoy-<ver>-linux.tar.gz`). Building from source is impractical: Ventoy bundles and compiles GRUB2 + EDK2 (UEFI) + IPXE using vendored cross-toolchains (Linaro/bootlin/mips/musl) in a CentOS 7.8 env — upstream itself builds it via Docker. Every distro (CachyOS, AUR) uses the prebuilt release; this mirrors the CachyOS `ventoy-bin` packaging.
- Pattern follows CachyOS: install to `/opt/ventoy/` (hardcoded in upstream scripts), wrappers in `%{_bindir}` (`ventoy`, `ventoygui`, `ventoyweb`, `ventoyplugson`, `ventoy-persistent`, `ventoy-extend-persistent`), `.desktop` + icon, system `xzcat`/`hexdump` symlinked in, bundled tool `.xz` binaries decompressed at build time, log path → `/var/log/ventoy.log`, shebangs → bash, and a `sanitize.patch` (the version-drifted `Ventoy2Disk.sh` hunk was dropped — its logic is self-consistent and covered by the build-time decompress).
- The `.desktop` launches `ventoygui` as the current user; VentoyGUI escalates to root internally via **polkit** (system password prompt) — `polkit` is in `Recommends`. No `sudo` in the `.desktop`.
- `ExclusiveArch: x86_64 aarch64`; `__strip /bin/true` + `debug_package %nil` (prebuilt binaries); `update.rhai` via `gh("ventoy/Ventoy")`.
- Tested: `anda build -c terra-rawhide-x86_64` → green; installed in `fedora:rawhide` — wrappers, `VentoyGUI.<arch>`, `boot.img`, `.desktop` and icon all land correctly.

**Checklist**

## Distrobox test (with clean Fedora image)
image

### DIFF
--- a/anda/tools/ventoy/anda.hcl
+++ b/anda/tools/ventoy/anda.hcl
@@ -1,0 +1,5 @@
+project pkg {
+	rpm {
+		spec = "ventoy.spec"
+	}
+}
--- a/anda/tools/ventoy/sanitize.patch
+++ b/anda/tools/ventoy/sanitize.patch
@@ -1,0 +1,81 @@
+Author: DuckSoft <realducksoft@gmail.com>
+
+I just cannot stand such a dirty package. 
+Let me do some optimisations.
+
+If you are interested to improve this patch, welcome to mail me.
+
+diff -Npur a/tool/ventoy_lib.sh b/tool/ventoy_lib.sh
+--- a/tool/ventoy_lib.sh	2026-04-05 09:27:04.000000000 -0600
++++ b/tool/ventoy_lib.sh	2026-04-06 10:03:00.691008472 -0600
+@@ -51,31 +51,6 @@ vtoy_gen_uuid() {
+ }
+ 
+ check_tool_work_ok() {
+-    
+-    if echo 1 | hexdump > /dev/null; then
+-        vtdebug "hexdump test ok ..."
+-    else
+-        vtdebug "hexdump test fail ..."
+-        ventoy_false
+-        return
+-    fi
+-   
+-    if mkexfatfs -V > /dev/null; then        
+-        vtdebug "mkexfatfs test ok ..."
+-    else
+-        vtdebug "mkexfatfs test fail ..."
+-        ventoy_false
+-        return
+-    fi
+-    
+-    if vtoycli fat -T; then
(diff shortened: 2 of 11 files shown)


Owen-sz:
We've discussed adding Ventoy before, but due to questionable binary blobs included in the source code we determined that we cannot be sure there is not a security risk with this package. More info can be found here - https://github.com/NixOS/nixpkgs/issues/404663
I appreciate the PR though, and am happy to discuss further.

c4i0b:
Thanks for the context @Owen-sz. We weren't aware of the blob transparency concerns or the NixOS issue when putting this together. That's a fair position. Closing the PR. Happy to revisit if upstream ever addresses the blob situation.

Owen-sz:
> Thanks for the context @Owen-sz. We weren't aware of the blob transparency concerns or the NixOS issue when putting this together. That's a fair position. Closing the PR. Happy to revisit if upstream ever addresses the blob situation.

Sounds good, I'd also love to have this package if upstream can provide transparency.

PR #14730, opened 2026-07-30
add: desktop-plus

c4i0b:

**Desktop Plus** — a community fork of GitHub Desktop with advanced functionality (MIT, Electron).

- Source: https://github.com/desktop-plus/desktop-plus
- Website: https://desktop-plus.org
- License: MIT

**Describe the motivation**

Not in Fedora or Terra. The official GitHub Desktop has no Linux builds; Desktop Plus is the actively-maintained fork that ships Linux RPM/DEB/AppImage releases (~594★). Features include commit search, multi-account (GitHub/Bitbucket/GitLab/Codeberg), commit graph, and multiple stashes per branch.

**Additional context**

Repackages the upstream **Linux RPM release** (`DesktopPlus-v<ver>-linux-<arch>.rpm`), the same approach `cursor` takes with its upstream package. A source build was investigated but is impractical here: provider sign-in needs OAuth client secrets (×4) that are GitHub Actions secrets and not in the repo, and the upstream pipeline uses electron-packager + electron-installer-redhat rather than stock electron-builder. The prebuilt RPM embeds those credentials, so repackaging it (as AUR `desktop-plus-bin` does) is the pragmatic choice.

Since `rpm2cpio` extraction discards the upstream RPM's `postinstall` scriptlet, the things it handled are reproduced at build time:

- **libcurl-gnutls → libcurl.so.4**: the bundled git/copilot ELFs link `libcurl-gnutls.so.4` (the Debian split), which Fedora doesn't ship. The SONAME is rewritten in `%install` (the same `sed` the upstream scriptlet runs), and the residual `CURL_GNUTLS_3` versioned-symbol require is filtered via `__requires_exclude` — it's unsatisfiable in RPM metadata but only a harmless runtime warning, since the symbol still binds to Fedora's libcurl.
- **No setuid chrome-sandbox**: dropped for security (a setuid-root helper from a third-party prebuilt binary is a local-privilege-escalation surface). Fedora enables unprivileged user namespaces by default, so Electron's namespace sandbox stays fully active with no setuid binary — matching Terra's own `electron` package. Verified the app launches sandboxed as a non-root user.
- `desktop-plus-cli` symlink recreated (upstream built it in the lost scriptlet).
- `Recommends: (gnome-keyring or kf6-kwallet)` for the Secret Service the app uses via `libsecret` (auto-required); KF5 dropped as EOL on rawhide.
- `ExclusiveArch: x86_64 aarch64`; `update.rhai` via `gh("desktop-plus/desktop-plus")`.

Tested: `anda build -c terra-rawhide-x86_64` → green; aarch64 built and install-tested locally too. Both install cleanly in `fedora:rawhide` and the app launches.

**Checklist**

### DIFF
--- a/anda/apps/desktop-plus/anda.hcl
+++ b/anda/apps/desktop-plus/anda.hcl
@@ -1,0 +1,5 @@
+project pkg {
+	rpm {
+		spec = "desktop-plus.spec"
+	}
+}
--- a/anda/apps/desktop-plus/desktop-plus.spec
+++ b/anda/apps/desktop-plus/desktop-plus.spec
@@ -1,0 +1,55 @@
+%undefine __brp_mangle_shebangs
+%define debug_package %{nil}
+%global __strip /bin/true
+%global _build_id_links none
+
+%ifarch x86_64
+%global rpmarch x86_64
+%elifarch aarch64
+%global rpmarch arm64
+%endif
+
+Name:           desktop-plus
+Version:        3.6.4.1
+%electronmeta
+Release:        1%{?dist}
+Summary:        A GitHub Desktop fork with advanced functionality and improvements
+License:        MIT
+URL:            https://desktop-plus.org
+Source0:        https://github.com/desktop-plus/desktop-plus/releases/download/v%{version}/DesktopPlus-v%{version}-linux-%{rpmarch}.rpm
+Packager:       Caio Bruno <cbrunofb@gmail.com>
+
+ExclusiveArch:  x86_64 aarch64
+
+BuildRequires:  cpio
+Recommends:     (gnome-keyring or kf6-kwallet or kf5-wallet)
+
+%description
+Desktop Plus is a community fork of GitHub Desktop with additional features:
+commit search, multi-account support (GitHub, Bitbucket, GitLab, Codeberg),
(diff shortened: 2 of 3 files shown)


c4i0b:
Going to close this and revisit. I attempted building from source (the project is MIT/open-source), but the electron-builder pipeline doesn't complete in a mock/rpmbuild environment yet. I'd prefer to ship a source-built package rather than repackaging the upstream release RPM. Will reopen when the build is sorted. Appreciate it!

Owen-sz:
> Going to close this and revisit. I attempted building from source (the project is MIT/open-source), but the electron-builder pipeline doesn't complete in a mock/rpmbuild environment yet. I'd prefer to ship a source-built package rather than repackaging the upstream release RPM. Will reopen when the build is sorted. Appreciate it!

Sounds good. What was the error? We may be able to help.

c4i0b:
Thanks @Owen-sz! I couldn't reopen this PR after a force-push on the branch, so I've opened a new one here: #14789. In short, I ran into some friction building the Electron app from source, so it now repackages the upstream Linux RPM instead. Could you take a look at #14789? Appreciate the help!

PR #15010, opened 2026-08-07
fix(apparmor): package systemd unit, fix profile/parser file-list conflict, bump release

CatPieLeaf:
## What got fixed:

- `apparmor.service`, `/usr/lib/apparmor/*`, and `aa-teardown` are `%dnl`'d
  out of `%files parser` — `init/` is never built, so the systemd unit is
  never installed. Nothing ever loads profiles into the kernel at boot as a
  result, and anything execing with `AppArmorProfile=` (e.g. dbus-broker's
  systemd drop-in) fails outright because the profile it expects was never
  loaded.
- `%files parser` also carries a bare `%{_sysconfdir}/apparmor.d/` line.
  Listing a directory path without `%dir` makes RPM own it *recursively*,
  which duplicates every path `%files profiles` already claims by name/glob.
  Installing both subpackages together hard-conflicts on any shared file
  (e.g. `usr.sbin.dnsmasq`), and the `# FIXME: the confusion...? how did
  this happen` comment already sitting on that line suggests this has been
  a known-but-unresolved landmine.
- `%files profiles`' enumerated file list is stale against upstream's actual
  bundled profile set — 5.0.2 ships roughly 150 sample profiles, not the
  dozen this list hardcodes (php-fpm, samba-*, zgrep, bin.*/sbin.*/usr.*,
  lsb_release, nvidia_modprobe). It was only "working" because the blanket
  parser-package line above was silently absorbing everything the profiles
  list missed.

## What this changes

- Add the `init/` `%make_install` step (`DISTRO=redhat`, wiring
  `APPARMOR_BIN_PREFIX`/`SBINDIR`/`SYSTEMD_UNIT_DIR`) and un-dnl the
  `apparmor.service`/`aa-teardown`/`/usr/lib/apparmor` entries in
  `%files parser`.
- Remove the bare `%{_sysconfdir}/apparmor.d/` line from `%files parser`.
- Replace `%files profiles`' stale enumerated list with a single recursive
  `%config(noreplace) %{_sysconfdir}/apparmor.d/*` glob, so it can't drift
  out of sync with upstream again.
- Enable parser cache write/compression by default, with a dedicated
  `/var/cache/apparmor` cache-loc.
- Fix `LibAppArmor`'s moved-into-a-subdir `__init__.py` — without it,
  `import LibAppArmor` silently returns an empty namespace package instead
  of the real SWIG bindings.
- `utils` subpackage: `Requires: apparmor-parser`; `Obsoletes` the
  `setroubleshoot` family (no purpose once AppArmor is the active LSM).
- Add two more upstream sample-profile patches: dnsmasq nsswitch denial fix,
  waydroid denial fix.
- Bump `Release` so these fixes actually land on reinstall instead of being
  skipped as a same-NVR no-op.
- Add `apparmor-libs32` (new package dir, `anda/lib/apparmor-libs32/`): a
  trimmed i686 multilib companion carrying just `libraries/libapparmor` (the
  shared lib + devel headers, no python bindings/parser/utils/pam/httpd
  module) so 32-bit consumers (Steam/Proton/Wine) can link against
  `libapparmor.so.1` on x86_64 systems. Ported from Rakuos's
  `apparmor-libs32.spec`, adapted to Terra's version (5.0.2) and Anda's
  native per-arch build mechanism (`arches = ["x86_64", "i686"]`) instead of
  a hand-rolled mock target_arch build. `update.rhai` mirrors the main
  package's so both stay version-synced automatically.
- Patch the SWIG `%new_copy_array` removal (see "CI failure" below) so the
  package actually builds on Terra's rawhide CI, which already carries
  swig-4.5.0.

## CI failure: SWIG `%new_copy_array` removed in swig 4.5.0

Terra's rawhide job (`anda/lib/apparmor/pkg`, x86_64 + aarch64) failed at
`%build` in the python SWIG bindings:

```
libapparmor_wrap.c: In function '_wrap_aa_splitcon':
libapparmor_wrap.c:6601:14: error: expected expression before '%' token
      arg1 = %new_copy_array(con_ptr, con_len+1, char);
```

Not a bug in this package's sources - builds cleanly against swig-4.4.1. SWIG 4.5.0 (what
Terra's rawhide builder has) deleted the `%new_copy_array` macro outright
(documented in upstream SWIG's `CHANGES.current`, 2026-05-16 entry by
erezgeva as a "potentially incompatible change"); calls to it now pass
through unexpanded into the generated C, which is invalid syntax. apparmor
upstream's `master` branch still calls it unchanged as of this writing, so
this isn't fixed upstream yet either. Added
`0001-fix-swig-new_copy_array-removed-in-swig-4.5.0.patch`, replacing the
one call site (in the `aa_splitcon` typemap) with the manual malloc+memcpy
equivalent, using the same `SWIG_exception_fail`/`SWIG_MemoryError`
allocation-failure handling already used elsewhere in that same interface
file.

### DIFF
--- a/anda/lib/apparmor/apparmor.spec
+++ b/anda/lib/apparmor/apparmor.spec
@@ -7,7 +7,7 @@
 
 Name:           apparmor
 Version:        5.0.2
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        AppArmor userspace components
 
 %define baseversion %(echo %{version} | cut -d. -f-2)
@@ -18,6 +18,8 @@ URL:            https://gitlab.com/apparmor/apparmor
 Source0:        %url/-/archive/v%normver/apparmor-v%normver.tar.gz
 Source1:        apparmor.preset
 Patch01:        0001-fix-avahi-daemon-authselect-denial-in-fedora.patch
+Patch02:        0001-fix-denial-on-dnsmask-for-nsswitch.patch
+Patch03:        0001-fix-apparmor-waydroid-denials.patch
 
 BuildRequires:  gcc
 BuildRequires:  automake
@@ -116,6 +118,12 @@ program profiles to the AppArmor Security kernel module.
 Summary:        AppArmor User-Level Utilities
 Requires:       python3-apparmor = %{version}
 Requires:       python3-notify2
+Requires:       %{name}-parser
+# SELinux troubleshooting tools have no equivalent purpose once AppArmor is
+# the active LSM, and aa-notify/aa-status already cover the same ground.
+Obsoletes:      setroubleshoot
+Obsoletes:      setroubleshoot-server
+Obsoletes:      setroubleshoot-plugins
 
 %description    utils
 This package provides the aa-logprof, aa-genprof, aa-autodep,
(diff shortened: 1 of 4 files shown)


anda/lib/apparmor/apparmor.spec
```diff
@@ -3,21 +3,28 @@
 %global __arch_install_post /bin/true
 %global _sbindir /usr/sbin
 
+# Rakuos's fork hit a debuginfo Name collision with apparmor-libs32.spec
+# on EL10 CI. Disabled defensively here too.
+%global debug_package %{nil}
```
Owen-sz, line 10: Could we just disable this on el10 if it works on Fedora branches?

Owen-sz (changes requested):

anda/lib/apparmor/apparmor.spec
CatPieLeaf: Done

Owen-sz:
Could you add a changelog entry too?

Owen-sz (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|el10|[](https://github.com/terrapkg/packages/pull/15059)|
|✅|f43|[](https://github.com/terrapkg/packages/pull/15060)|
|✅|f44|[](https://github.com/terrapkg/packages/pull/15061)|
|✅|f45|[](https://github.com/terrapkg/packages/pull/15062)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/31227811256) for details

PR #15132, opened 2026-08-09
fix: protonplus

Owen-sz:

PR #15155, opened 2026-08-09
fix: protonplus

halfcyan:
idk why it grabbed the version wrong? it might also continue trying to grab the version incorrectly but it should build from this at least once lol


Owen-sz:
RPM versions can't have a `-`

It pulled the version right, upstream pushed a hotfix release. I have a fix I'll push soon to https://github.com/terrapkg/packages/pull/15132

PR #15366, opened 2026-08-12
chore (mwc): use %conf

Owen-sz:


Owen-sz:
This package hasn't been updated in over a year, and needs an older scenefx version. I think we should just remove this.

PR #15442, opened 2026-08-13
fix: Anki version typo

struggle387:

PR #15443, opened 2026-08-13
fix(anki-bin): update sources and fix versioning

madonuko:


madonuko:
hmct changelog update sources and versioning

madonuko:
goto #15442

PR #16214, opened 2026-08-26
remove: noctalia

halfcyan:
specifically removes noctalia stable as it's been packaged in Fedora now (https://packages.fedoraproject.org/pkgs/noctalia/noctalia/)


halfcyan:
will require satm remove

Owen-sz (approved):

PR #16248, opened 2026-08-27
add: unrar

ammix:
Closes #16139

Hi! This package is loosely based on RPM Fusion’s unrar spec.

### Notes

- Includes the Debian man page. This adds some maintenance burden, so I am happy to drop it if it is not desired.
- Does not include libunrar for now. I can add it here or in a follow-up PR if preferred/interested.

The current SPDX identifier for the UnRAR license is UnRAR, but Fedora’s license database does not include it yet, which causes rpmlint to report it as invalid. I have used the older LicenseRef-unrar identifier for now, although I am not sure if this is the preferred approach.

PR #16257, opened 2026-08-27
add: libunrar and libunrar-devel

ammix:
Follow-up to #16248

### DIFF
--- a/anda/tools/unrar/unrar.spec
+++ b/anda/tools/unrar/unrar.spec
@@ -1,6 +1,6 @@
 Name:           unrar
 Version:        7.2.7
-Release:        1%{?dist}
+Release:        2%{?dist}
 Summary:        Utility for extracting, testing and viewing RAR archives
 
 License:        LicenseRef-unrar AND BSD-2-Clause AND CC0-1.0 AND LicenseRef-Fedora-Public-Domain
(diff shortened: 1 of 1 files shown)


anda/tools/unrar/unrar.spec
```diff
@@ -18,26 +18,68 @@ BuildRequires:  make
...
 %{_bindir}/unrar
 %{_mandir}/man1/unrar.1*
 
+%files -n libunrar
+%license license.txt acknow.txt
+%doc readme.txt
+%{_libdir}/libunrar.so
+
+%files -n libunrar-devel
+%license license.txt acknow.txt
```
Owen-sz: You don't need to install the license files on subpackages which dep on base packages with license files.

Owen-sz:
If you want, could you try using `%pkg_devel_files`?

https://docs.terrapkg.com/contributing/srpm/#pkg_devel_files

anda/tools/unrar/unrar.spec
ammix: Oh I see. I once again just followed RPM Fusion here. Fixed in c4ea41a

ammix:
Yeah, I saw `%pkg_devel_files`, but upstream only builds an unversioned `libunrar.so`. In this case, that file is the actual runtime library rather than a development linker symlink, while the macro includes `%{_libdir}/*.so` in the devel package.

I could use `%pkg_devel_files -n libunrar-devel -N libunrar` with `%exclude %{_libdir}/libunrar.so`, then keep the library explicitly under `libunrar`. Would you prefer that, or keeping the explicit list here?

Owen-sz:
> Yeah, I saw `%pkg_devel_files`, but upstream only builds an unversioned `libunrar.so`. In this case, that file is the actual runtime library rather than a development linker symlink, while the macro includes `%{_libdir}/*.so` in the devel package.
> 
> I could use `%pkg_devel_files -n libunrar-devel -N libunrar` with `%exclude %{_libdir}/libunrar.so`, then keep the library explicitly under `libunrar`. Would you prefer that, or keeping the explicit list here?

Ok, yeah this is a special case. We can leave it as is. Thanks!

Owen-sz (approved):

raboneko:

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|el10|[](https://github.com/terrapkg/packages/pull/16258)|
|✅|f43|[](https://github.com/terrapkg/packages/pull/16259)|
|✅|f44|[](https://github.com/terrapkg/packages/pull/16260)|
|✅|f45|[](https://github.com/terrapkg/packages/pull/16261)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport) and see the [Github Action logs](https://github.com/terrapkg/packages/actions/runs/33107667945) for details

Issue #16665, opened 2026-09-07
[BUG] terra-gamescope never upgrades past 137.7282613d because the git-hash Version doesn't sort


exformation:
### Full Raw Package Name (e.g. anda-0.4.14-1.fcrawhide.x86_64)

terra-gamescope-137.77e6ea94-1.fc44.x86_64

### Release Version

f44

### Describe the Bug

terra-gamescope's Version is `137.<short commit>`, and rpm reads the hash's leading digits as a number, so the July build (137.7282613d) outranks every pin since, including #16498's 137.77e6ea94. dnf keeps installing July's build, and Bazzite images have shipped 137.7282613d-1 on every release since testing-44.20260726.

### To Reproduce

dnf5 -q repoquery --disablerepo='*' --repofrompath=tx,https://repos.fyralabs.com/terra44-extras --enablerepo=tx --arch x86_64 --latest-limit=1 terra-gamescope

### Expected Behavior

What it prints today:
terra-gamescope-0:137.7282613d-1.fc44.x86_64
  built 2026-07-25, gamescope code authored 2026-05-08

What it should print, if versions sorted by age:
terra-gamescope-0:137.77e6ea94-1.fc44.x86_64
  built 2026-09-04, gamescope code from 2026-08-24

### Relevant Log Output

```shell

```

PR #16666, opened 2026-09-07
chore (terra-gamescope): proper versioning

Owen-sz:


Owen-sz:
Should fix https://github.com/terrapkg/packages/issues/16665

PR #16754, opened 2026-09-10
fix(bazel): remove large runner label

halfcyan:


Owen-sz:
Should this be closed?

PR #16854, opened 2026-09-11
bolt-launcher fixes

TheHeroBrine422:
- fix bolt building due to bad cef path
- move bolt to bolt-launcher so folder/spec name match name in the spec.


TheHeroBrine422:
something went wrong with this PR. ignore it.

Issue #16726, opened 2026-09-08
[BUG] `spotify-launcher-0.6.6-1.fc44` requires `compat-ffmpeg4` to play local M4A files.


beedell-jared:
### Full Raw Package Name (e.g. anda-0.4.14-1.fcrawhide.x86_64)

spotify-launcher-0.6.6-1.fc44.x86_64

### Release Version

f44

### Describe the Bug

When attempting to play local files the error "Spotify can't play this right now" shows, and the song doesn't play. I fixed this issue by installing `compat-ffmpeg4`, as Spotify requires older versions of FFmpeg, which Fedora does not include. Please make this a dependency.

(Also, it should include `libayatana-appindicator`, per the logs.)

### To Reproduce

1. Run spotify-launcher
2. Enable local songs in settings
3. Play any local song that is a m4a file

### Expected Behavior

I expected the song to play as with streaming/downloaded songs.

### Relevant Log Output

```shell
Red@Beedell:~$ /usr/bin/spotify-launcher
[2026-09-08T21:51:30Z INFO  spotify_launcher::config] Loading configuration file at "/etc/spotify-launcher.conf"
[2026-09-08T21:51:30Z INFO  spotify_launcher] No update needed

(spotify:20093): libayatana-appindicator-WARNING **: 22:51:30.315: libayatana-appindicator is deprecated. Please use libayatana-appindicator-glib in newly written code.
```

Owen-sz:
I fixed the libayatana-appindicator issue, but we cannot dep on a package not in Terra/Fedora (compat-ffmpeg4 comes from RPM Fusion).

beedell-roke:
@Owen-sz, would adding that package be worth a separate issue, then?

Owen-sz:
> @Owen-sz, would adding that package be worth a separate issue, then?

Yeah, I wonder if this issue can be fixed by installing any of the fedora ffmpeg packages though?

ammix:
> > [@Owen-sz](https://github.com/Owen-sz), would adding that package be worth a separate issue, then?
> 
> Yeah, I wonder if this issue can be fixed by installing any of the fedora ffmpeg packages though?

Don’t think that’s possible with the Fedora 44+ packages currently available. There's no version providing any of the ABIs supported by Spotify [(61 and under)](https://repository.spotify.com/dists/stable/non-free/binary-amd64/Packages). ffmpeg 8 and 9 provide ABIs 62 and 63 and theres no compat package.

[...]

PR #16935, opened 2026-09-12
fix(spotify-launcher): add optional ffmpeg dependency

ammix:
Follow up to the conversation in #16726

Just adding an optional dependency that will pull in compat-ffmpeg4 from rpmfusion (tested working), or a possible future terra compat-ffmpeg package later. If you want a compat package in terra, I already have a compat-ffmpeg7 spec nearly ready to go, which is just a minimal version of the current ffmpeg spec. I can open a PR for that too if wanted.


Owen-sz:
Thanks! Yeah unfortunately we cannot dep on packages not in Terra or Fedora, so I'd mark this as a draft for now until `compat-ffmpeg7` is added.
And in that case, we could switch the `Requires:` to the package name instead of `.so` names?

ammix:
Makes sense. My thinking with the `.so` recommends was just to keep the optional dependency flexible, so rpmfusions compat-ffmpeg4 could satisfy it too for people using terra without terra-multimedia. But yeah, I'll switch it over to compat-ffmpeg7 once that's merged.

beedell-roke:
> unfortunately we cannot dep on packages not in Terra or Fedora

@Owen-sz, why? I ask because @ammix's approach <em>should</em> allow it to install, if RPMFusion is available, considering that DNF5 is able to cope with resolving dependencies across repositories, insofar as a `.so` is listed as required, and solely fulfilled by another repository's package[s].

Owen-sz:
> Makes sense. My thinking with the `.so` recommends was just to keep the optional dependency flexible, so rpmfusions compat-ffmpeg4 could satisfy it too for people using terra without terra-multimedia. But yeah, I'll switch it over to compat-ffmpeg7 once that's merged.

Sounds good, yeah makes sense. I think we should still use the package name once the package is merged in case the `.so` files change names in the future.

Owen-sz:
> > unfortunately we cannot dep on packages not in Terra or Fedora
> 
> @Owen-sz, why? I ask because @ammix's approach _should_ allow it to install, if RPMFusion is available, considering that DNF5 is able to cope with resolving dependencies across repositories, insofar as a `.so` is listed as required, and solely fulfilled by another repository's package[s].

Good question, there are distros that ship Terra that do not ship RPM Fusion, and this could break those.

ammix:
> > > unfortunately we cannot dep on packages not in Terra or Fedora
> > 
> > 
> > @Owen-sz, why? I ask because @ammix's approach _should_ allow it to install, if RPMFusion is available, considering that DNF5 is able to cope with resolving dependencies across repositories, insofar as a `.so` is listed as required, and solely fulfilled by another repository's package[s].
> 
> Good question, there are distros that ship Terra that do not ship RPM Fusion, and this could break those.

Since its only a `Recommends:` I don't think an unavailable provider should break installation. Using the package name is easier to maintain though, so I'm happy to switch it over.

One thing I noticed just now though: `compat-ffmpeg7` would conflict with regular ffmpeg on Fedora 43 and EL10. Should we just skip the backports there and keep the `.so` recommendation so fedoras ffmpeg can satisfy it, or just base the compat package on ffmpeg 6 so it could be used anywhere?

Owen-sz:
> > > > unfortunately we cannot dep on packages not in Terra or Fedora
> 
> > > 
> 
> > > 
> 
> > > @Owen-sz, why? I ask because @ammix's approach _should_ allow it to install, if RPMFusion is available, considering that DNF5 is able to cope with resolving dependencies across repositories, insofar as a `.so` is listed as required, and solely fulfilled by another repository's package[s].
> 
> > 
> 
> > Good question, there are distros that ship Terra that do not ship RPM Fusion, and this could break those.
> 
> 
> 
> Since its only a `Recommends:` I don't think an unavailable provider should break installation. Using the package name is easier to maintain though, so I'm happy to switch it over.
> 
> 
> 
> One thing I noticed just now though: `compat-ffmpeg7` would conflict with regular ffmpeg on Fedora 43 and EL10. Should we just skip the backports there and keep the `.so` recommendation so fedoras ffmpeg can satisfy it, or just base the compat package on ffmpeg 6 so it could be used anywhere?

Hmm, yeah I wonder if doing something like `Recommends: (file.so or compat-ffmpeg7)` would be best here.

PR #16942, opened 2026-09-12
fix: Drop back to libclc until libclc22 is rolled out everywhere

KyleGospo:

PR #16971, opened 2026-09-13
fix(mesa): require libclc 22

adil192:
Since the separate `libclc22` package isn't available yet (https://github.com/terrapkg/packages/pull/16942), we should explicitly require libclc version 22.

libclc has already updated to version 23 at least on Fedora 45 updates-testing, so we should prevent that update. Fedora 44 and older aren't affected.

In the future, this should be updated to use the `libclc22` package: https://bodhi.fedoraproject.org/updates/FEDORA-2026-06a5023802.

Mirrors a change in my copr repo: https://github.com/adil192/mesa-packaging/commit/d4a763a7900f7f797ed79dd52856e9a3a6bb005a


adil192:
Closing since this only matters on f45 and newer, and terra45-mesa is failing to build right now due to missing mirrors/metalink.
Once builds succeed again, libclc22 will likely have already been released

